PR #18, opened 2015-03-01
Travis-CI changes and build improvements

zmughal:
The branch `travis-ci-more` makes several changes:

### View code:

- Sourceforge: <https://sourceforge.net/p/pdl/code/ci/travis-ci-more/tree/>
- GitHub: <https://github.com/PDLPorters/pdl/pull/18>

### Summary:

1. @mohawk2 has worked on incorporating the code for the pre-req report from [Dist::Zilla::Plugin::Test::ReportPrereqs](http://p3rl.org/Dist::Zilla::Plugin::Test::ReportPrereqs) into a test that can be used to dump the versions of the pre-reqs used. This also helps with finding the versions of the packages during the Travis-CI build.
2. There is an additional Travis-CI build which skips installing `gfortran` so that PDL can be tested in the absence of a Fortran compiler. This addresses the issue [SF#346 ExtUtils::F77 dependency causing problems for CPAN install](https://sourceforge.net/p/pdl/bugs/346/).
3. @mohawk2 made fixes the tests so that they behave properly when `WITH_BADVAL=0` and all the corresponding Travis-CI builds now succeed.
4. Travis-CI now gives IRC notifications to `irc.perl.org/#pdl` when the build succeeds or fails.
5. Several changes including parallel builds and testing so that Travis-CI finishes faster. This speeds up the overall build by 25%.

Code review thread on pdl-devel: <http://thread.gmane.org/gmane.comp.lang.perl.pdl.devel/6328>

PR #21, opened 2015-03-01
Use Alien::HDF4 in PDL::IO::HDF build

mohawk2:
This is for discussion; the first use of a real, working Alien module in the PDL build process. Note that it uses `ExtUtils::Depends` and therefore adds it as a dep.

Consider it a fully-working prototype for considerably tidying up the rest of the build system.


zmughal:
Regarding the failed build, this is only failing because #18 hasn't been merged yet.

mohawk2:

PR #22, opened 2015-03-02
remove Devel::CheckLib from inc/

zmughal:
We already have `Devel::CheckLib` in the `Makefile.PL` under
`CONFIGURE_REQUIRES`.

This fixes <https://sourceforge.net/p/pdl/bugs/374/>.


zmughal:
Regarding the failed build, this is only failing because #18 hasn't been merged yet.

mohawk2:

PR #51, opened 2015-03-02
Add propagation of badflag with .= (Ops::assgn) for sf.net bug 3543056.

mohawk2:
Rebased version of #2. Passes tests.

### DIFF
--- a/Basic/Ops/ops.pd
+++ b/Basic/Ops/ops.pd
@@ -407,14 +407,26 @@ sub PDL::log10 {
 #
 pp_def(
        'assgn',
-#       HandleBad => 1,
+       HandleBad => 1,
        Pars => 'a(); [o]b();',
        Code =>
        '$b() = $a();',
-#       BadCode =>
-#       'if ( $ISBAD(a()) ) { $SETBAD(b()); } else { $b() = $a(); }',
+       BadCode =>
+       'if ( $ISBAD(a()) ) { $SETBAD(b()); } else { $b() = $a(); }',
        Doc =>
 'Plain numerical assignment. This is used to implement the ".=" operator',
+		BadDoc =>
+'If C<a> is a child piddle (e.g., the result of a slice) and bad values are generated in C<b>,
+the bad value flag is set in C<b>, but it is B<NOT> automatically propagated back to the parent of C<a>.
+The following idiom ensures that the badflag is propagated back to the parent of C<a>:
+
+ $pdl->slice(":,(1)") .= PDL::Bad_aware_func();
+ $pdl->badflag(1);
+ $pdl->check_badflag();
+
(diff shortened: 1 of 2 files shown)


zmughal:
I believe this needs to only run when `WITH_BADVAL=1`. I don't think the test does so yet.

mohawk2:
We'll be able to tell more easily once #18 is merged and we rebase against it.

mohawk2:

mohawk2:

zmughal:
LIke I was saying in IRC <http://irclog.perlgeek.de/pdl/2015-03-03#i_10215431>, this won't fix the code until the `badflag` does not need to be set. Turning into TODO tests for now.

mohawk2:
@hainest, what are your thoughts on the readiness state of this? Does work remain to be done?

hainest:
I think that as long as folks are content with the lack of child->parent bad flag propagation, then this can be pulled into master. The (admittedly verbose) idiom for ensuring bad flag propagation is provided in the comments of assign.

mohawk2:
I think merging should wait until post 2.008, which is imminent. However, there is talk of a much faster release, so that'll help.

May I suggest changing the location of telling people the correct idiom move from comments, to actual documentation? Feel like making a PR onto this branch? I'll then bring it across and we can see how it plays.

hainest:
>May I suggest changing the location of telling people the correct idiom move from comments, to actual documentation?

I misspoke. The description is in the Doc field, so it should be rendered to the documentation automatically.

>Feel like making a PR onto this branch?

I thought it was already pulled. Github says that I can merge the PR, but I'm not sure how that aligns with the rebase procedure.

mohawk2:
It means this branch, as rebased by me, *can* be merged. It doesn't mean you can do so, until you get a commit bit on the appropriate repo (of which this is a mirror). If you make your fork be up to date with this repo, then branch off this branch, then PR onto this branch (or against master, I'll figure it out), then you can submit changes to your code here. That is what I am saying you could do, if you feel it useful.

mohawk2:
Rebased to update for `t/ops.t` change.

mohawk2:
@hainest @d-lamb @devel-chm @zmughal @drzowie Is this ready for primetime? :-)

hainest:
@mohawk2 I think so. I would reiterate my previous comment:

> I think that as long as folks are content with the lack of child->parent bad flag propagation, then this can be pulled into master. The (admittedly verbose) idiom for ensuring bad flag propagation is provided in the doc for `assign`.

mohawk2:
Limitations are documented. Tests are passing. Merging.

mohawk2:
The above-linked commit actually now means `assgn` does not need the `badflag` set on its output ndarray, which is nice to finally achieve.

PR #54, opened 2015-03-03
GitHub sprucing

zmughal:
Not much to see here, but this will do for now. PDL needs a pretty front gate on GitHub. Text is from the website.

selection_385


zmughal:
Updated to add badges.

selection_386

PR #63, opened 2015-03-07
descriptive warning when using the eq overloaded operator on non-numeric strings

zmughal:
Instead of saying "isn't numeric in null operation", provide a more
descriptive warning that indicates the value and the operator being
used.

See <http://sourceforge.net/p/pdl/bugs/332>,
<https://github.com/PDLPorters/pdl/issues/33> for more information.


mohawk2:
This looks good to me: I second it.

PR #66, opened 2015-03-13
Test manifest

mohawk2:


zmughal:
Nice! This will be very useful.

PR #68, opened 2015-03-14
Travis-CI: use perl Makefile.PL directly

zmughal:
The Travis-CI helper's `build-dist` command uses Perl 5.20 for running
every build run. This simulates a module author releasing a dist for
CPAN with `META.*` files.

By switching that out and running `Makefile.PL` directly, the run uses
the Perl version specific to that run, but it also becomes necessary to
install the `CONFIGURE_REQUIRES` requirements beforehand because the
`META.*` files are not generated yet.


zmughal:
Obsoleted by #71.

PR #71, opened 2015-03-14
Travis-CI: set SPLIT_BUILD to false to use the run's version of Perl for configure-time

zmughal:
By default, the Travis-CI helper's `build-dist` command uses Perl 5.20
for running every build run. This simulates a module author releasing a
dist for CPAN with `META.*` files.

By switching that out and running `Makefile.PL` directly, the run uses
the Perl version specific to that run, but it also becomes necessary to
install the `CONFIGURE_REQUIRES` requirements beforehand because the
`META.*` files are not generated yet.


zmughal:
If this passes, it obsoletes #68.

PR #73, opened 2015-03-15
"core" target fixes

mohawk2:
Currently the "core" target doesn't build `Math` or `MatrixOps` and both are needed. Also, using `gmake -j4`, the core SOs are usually (but not always) not built. This fixes or works around these issues.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -596,9 +596,19 @@ EOT
 
 my $basic_pmblib = $self->cd('Basic', "\$(MAKE) pm_to_blib");
 # the modules in PDL::LiteF, used in t/core.t
-my $basicbuild = join "\n\t",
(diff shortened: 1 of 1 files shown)


mohawk2:
Further pondering reminds me that the "core" target with the workaround only works for "dynamic". Likewise, the "coretest" only works with dynamic, although that's fixable with a recursive `$(MAKE) test TEST_FILES="t/core.t t/ops.t"`.

mohawk2:
(Although in "static", that recursive make would actually depend on "static" which would build the whole thing, thereby rendering the smaller build pointless).

zmughal:
This looks good to me. A little ugly, but as long as we get this working more nicely with future EU:MM releases, go for it!

PR #74, opened 2015-03-15
Make Filter::Util::Call the default engine for PDL::NiceSlice

devel-chm:
This is the original source filter implementation.  The more
correct one using Filter::Simple is too slow.  Work to fix is
underway but postponed for PDL-2.009

### DIFF
--- a/Basic/SourceFilter/NiceSlice.pm
+++ b/Basic/SourceFilter/NiceSlice.pm
@@ -5,9 +5,9 @@ BEGIN {
       'Module::Compile'     => 'PDL/NiceSlice/ModuleCompile.pm',
    );  # to validate names
 
-   $PDL::NiceSlice::engine = $engine_ok{'Filter::Simple'};  # default engine type
+   ## $PDL::NiceSlice::engine = $engine_ok{'Filter::Simple'};  # default engine type
    ## TODO: Add configuration argument to perldl.conf
-   ## $PDL::NiceSlice::engine = $engine_ok{'Filter::Util::Call'};  # default engine type
+   $PDL::NiceSlice::engine = $engine_ok{'Filter::Util::Call'};  # default engine type
 
    if ( exists $ENV{PDL_NICESLICE_ENGINE} ) {
       my $engine = $ENV{PDL_NICESLICE_ENGINE};


Basic/SourceFilter/NiceSlice.pm
```diff
@@ -5,9 +5,9 @@ BEGIN {
       'Module::Compile'     => 'PDL/NiceSlice/ModuleCompile.pm',
    );  # to validate names
 
-   $PDL::NiceSlice::engine = $engine_ok{'Filter::Simple'};  # default engine type
+   ## $PDL::NiceSlice::engine = $engine_ok{'Filter::Simple'};  # default engine type
    ## TODO: Add configuration argument to perldl.conf
-   ## $PDL::NiceSlice::engine = $engine_ok{'Filter::Util::Call'};  # default engine type
+   $PDL::NiceSlice::engine = $engine_ok{'Filter::Util::Call'};  # default engine type
```
mohawk2: It's a pity this is necessary for now, but evidently it is. We should probably add a test exercising the case in #25 / https://sourceforge.net/p/pdl/bugs/238/

mohawk2:
It's not 100% clear to me why the Filter::Simple version was so slow (the tests took about 900ms on my machine with that, vs about 220ms with the old one). However, inserting timing outputs showed MatrixOps was loading PDL::NiceSlice. After making that not be the case with https://github.com/PDLPorters/pdl/commit/6ae4c0e60e6c2bd96b95be800020327c1f0c6d4f, the difference went from about +300% to about +10%.

Given that the old version was broken with strings (including SQL), POD, and even comments with partial "nice slice" syntax, that is a price entirely worth paying. Therefore I have switched the default to the Filter::Simple with https://github.com/PDLPorters/pdl/commit/4a790d9a378de441ae4b305c0411ba97bf197914, which finally fixes this problem, and #25 and #32.

mohawk2:
The Filter::Simple implementation was in fact very slow on bigger files (such as the 6700-line PDL::LinearAlgebra taking 10s to load). Analysis revealed this was due to a performance bug in Text::Balanced, worked around in PDL with https://github.com/PDLPorters/pdl/commit/31b991b70c80ff9123757e746b13895343835bc6 and the same change submitted to Text::Balanced itself with https://github.com/steve-m-hay/Text-Balanced/pull/5

Issue #67, opened 2015-03-13
Create a hand-written Changes file


zmughal:
Using the `git log` is not a very high-level readable solution for changlogs. See <http://keepachangelog.com/>.

I volunteer to help maintain this file for each release.

zmughal:
This work is being initiated in the <https://github.com/PDLPorters/pdl/tree/changesfix> branch.

zmughal:
Also, make sure that the format of the Changes file is documented in `DEVELOPMENT` and part of the release process.

zmughal:
Closed via #78.

PR #78, opened 2015-03-15
Move to standard Changes format

zmughal:
This is to address #67.

This allows for the Changes file to be parsed on MetaCPAN by <https://metacpan.org/pod/CPAN::Changes>.


mohawk2:
Fantastic work!

xt/00-check-changelog.t
```diff
@@ -0,0 +1,17 @@
...
+use CPAN::Changes;
+use Data::Dumper;
+
+my $changes = CPAN::Changes->load('Changes');
+
+ok($changes);
+
+my @releases = map { +{ $_->version => $_->date } } $changes->releases;
+
+note Dumper \@releases;
```
mohawk2: I think if this were `diag` instead of `note` it would show up in the Travis log?

zmughal: It should still show up since I'm using `-v`, but turns out, it isn't running at all?

/me investigates.

mohawk2: I'll wager test-files might be confusing us. To the docs, Robin!

mohawk2: Need AUTHOR_TESTING set to true.

zmughal: Should default to true, but I'll make it explicit.

mohawk2:
Reviewed: looks great!

zmughal:
Sent to @PDLPorters <http://permalink.gmane.org/gmane.comp.lang.perl.pdl.devel/6502>

Issue #13, opened 2015-03-01
Travis-CI: add support for multiple compilers


zmughal:
PDL needs to be compiled with more than `gcc`. For a first start, `clang` should be tried. Later, look at Intel's compiler.

This will help for portability to OS X.

zmughal:
See the `clang_warnings` branch (PR in #84)

PR #85, opened 2015-03-18
Travis-CI: add build that uses Clang as CC compiler

zmughal:
This is in order to test against compilers other than GCC.

This fixes <https://github.com/PDLPorters/pdl/issues/13>.


zmughal:
While Travis-CI says that it passes, this is deceptive. Setting CC using the EUMM option does not work for subdirectory Makefiles.

    perl Makefile.PL CC=clang; find -name Makefile -exec grep -H '^CC =' {} \;

will only set the `CC=clang` in the top level Makefile. All others have `CC=cc`.

I have an ugly fix for this using symlinks, but I would rather fix the build.

mohawk2:
The way forward would be to configure that Perl installation with something like

```
sudo perl -pie "s/(cc => ')gcc(')/${1}clang$2/" /correct/path/Config.pm
```

PR #90, opened 2015-03-22
Parallelisable core target

mohawk2:
Before, "make core" took 2min45 on my 2-core system. WIth this, using GNU make -j2, it consistently takes about 2 mins.


mohawk2:
This was in fact merged and released with 2.018.

PR #91, opened 2015-03-29
Alien::Proj4 - fix include ordering error

mohawk2:


zmughal:
Looks good. Builds and tests pass on my machine.

mohawk2:
Great! @sisyphus says it works for him except for giving a false negative for `->installed`, so I'm working with him on that.

PR #98, opened 2015-04-10
Add Travis case for testing not blib but installed

mohawk2:

PR #99, opened 2015-04-10
minimalist inline pdlpp fix

drzowie:
Hopefully this one will work with Travis.


mohawk2:
After discussion with @drzowie, cherry-picked over to #98, which is now merged to master. Closing.

PR #109, opened 2015-05-24
Test cleanup 3

### LINKED ISSUE #8: Refactor to use Test::More
- [x] aaa_load.t
- [x] argtest.t
- [x] bool.t
- [x] clump.t
- [x] croak.t
- [x] familyfree.t
- [x] flow.t
- [x] foo.t
- [x] gauss.t
- [ ] hdrs.t
- [x] howbig.t
- [ ] ica.t
- [ ] imagend.t
- [ ] imagergb.t
- [ ] interp.t
- [ ] interp_slatec.t
- [ ] interpol.t
- [ ] limits_normalize_dsets.t
- [ ] linfit.t
- [ ] magic.t
- [ ] matmult.t
- [ ] ones.t
- [x] ops.t
- [ ] pdlchar.t
- [ ] physical.t
- [ ] picnorgb.t
- [ ] picrgb.t
- [ ] pnm.t
- [ ] poly.t
- [ ] polyroots.t
- [ ] primitive2.t
- [ ] pthread.t
- [ ] pthread_auto.t
- [ ] reduce.t
- [ ] refs.t
- [ ] scope.t
- [ ] segfault.t
- [ ] thread.t
- [ ] thread_def.t
- [ ] vaffine.t

zmughal:
Resolves #8, #9.

* using `Test::More` module (no hand-rolled TAP or `Test` module)
* `use strict; use warnings` everywhere
* remove use of `$a`, `$b` lexicals (reserved for `sort` function)
* using `Test::Exception` (lives_ok, dies_ok, etc.) rather than the `eval { $code }; ok $@` combination
* make each individual test run in its own scope
  e.g.

  ```perl
  {
	my $test = ...
	ok( ... );
  }
  {
	my $test = ...
	ok( ... );
  }
  ```

  This makes tests not dependent on previous code which means that they can be
  moved around without worrying about state. In the future, these should be setup
  as labelled subtests.
* Using `all approx( ... )` instead of a custom `tapprox( ... )` in each file.


mohawk2:
I've taken the liberty of squashing together the commits that had the same label ie were on the same file.

t/familyfree.t
```diff
@@ -4,16 +4,19 @@ use strict;
...
+	my $pb = $pa->slice(':,2:3');
 
-$pa = 1;  # Undefine orig. a
-$pb += 1;
+	$pa = 1;  # Undefine orig. a
+	$pb += 1;
+};
 
-ok(1);
+done_testing;
```
mohawk2: You've got both tests => 1, and `done_testing`. Pick one?

zmughal: Hmm, does having both help in case the test segfaults? I think you're right, I should remove it.

zmughal:
Sent to @PDLPorters <http://permalink.gmane.org/gmane.comp.lang.perl.pdl.devel/6502>

t/func.t
```diff
@@ -1,10 +1,13 @@
 # -*-perl-*-
 #
-use Test;
+use Test::More;
 use PDL::LiteF;
 
+use strict;
+use warnings;
+
+our $loaded; our $slatec;
```
zmughal: @mohawk2 pointed out on IRC that this should be `my` not `our`.

mohawk2:
t/callext.t - space vs tabs inconsistency
t/clump.t - just zap commented code
t/func.t - maybe done_testing? Also, L25 please avoid indirect object notation and use Class->init(...)
t/gauss.t - zap commented code
 t/hist.t - still has tapprox?
t/image2d.t - L39 onwards, please can we bite the bullet and indent at least these nested scopes? Hard to read otherwise

mohawk2:
t/imagergb.t - you have ok ($x == blah), why not is() or cmp_ok()?
t/lut.t - instead of using `$#array`, I much prefer eg `isnt scalar(@array), 0`
t/matrixops.t - what's wrong with `my $C22inv = eval { $C22->inv }`?

t/thread.t
```diff
@@ -22,62 +25,69 @@ ok("$b" eq <<END);
...
-maximum($a->thread(0,1),$c);
-# print "B:\n"; $b->dump(); print "C:\n";$c->dump();
-# print $b;
+maximum($pa->thread(0,1),$pc);
+# print "B:\n"; $pb->dump(); print "C:\n";$pc->dump();
+# print $pb;
 
-ok($b->at(0,0) == 10);
-ok($b->at(1,1) == 14);
+ok($pb->at(0,0) == 10);
```
mohawk2: Could we use `cmp_ok` here too?

mohawk2:
t/niceslice.t - L95 et seq: why not just `isnt $@, ''`? I'd zap all the `note` stuff, `is` will help us when needed.
t/simplex.t - L30 use `cmp_ok`?

t/transform.t
```diff
@@ -3,90 +3,107 @@ use warnings;
...
+{
 # Scaling by 2 in pixel coordinates should actually scale the image
-$b = $a->map(t_scale(2),{pix=>1});
-ok(all($b == $a*0.5));
+my $pb = $pa->map(t_scale(2),{pix=>1});
+ok(all($pb == $pa*0.5));
+}
+}
 
+{
```
mohawk2: Indent inside a block like this?

t/transform.t
```diff
@@ -102,60 +120,93 @@ ok(all(approx($m51_coords, $m51map_coords,1e-8)));
...
+my $pb = $pa->match($pa,{method=>'l'});
+ok(all(approx($pa,$pb)),"self-match with 'l' method is an approximate no-op");
+}
 
-$b = $a->match($a,{method=>'h'});
-ok(all(approx($a,$b)),"self-match wtih hanning method is an approximate no-op");
+{
+my $pb = $pa->match($pa,{method=>'h'});
+ok(all(approx($pa,$pb)),"self-match wtih hanning method is an approximate no-op");
```
mohawk2: wtih -> with

t/transform.t
```diff
@@ -102,60 +120,93 @@ ok(all(approx($m51_coords, $m51map_coords,1e-8)));
...
-ok(all(approx($b->clip(1e-9)->log10,$b0,1e-7)),"self-match with Gaussian method gives understood blur");
+ok(all(approx($pb->clip(1e-9)->log10,$b0,1e-7)),"self-match with Gaussian method gives understood blur");
+}
 
-$t = t_linear(pre=>[0.5,1]);
-$b = $a->map($t,{method=>'s',pix=>1});
-my $wndb = $b->whichND;
-ok($wndb->nelem==2 and all($wndb==pdl([[3,4]])) and approx($b->slice(3,4),1),'offset with sample is a simple offset');
+{
+my $t = t_linear(pre=>[0.5,1]);
```
mohawk2: Indent inside this long block to aid the poor reader?

t/xvals.t
```diff
@@ -1,17 +1,19 @@
...
-kill INT,$$ if $ENV{UNDER_DEBUGGER}; # Useful for debugging.
+kill 'INT',$$ if $ENV{UNDER_DEBUGGER}; # Useful for debugging.
 
 
-$a0 = zeroes(3,2);
+my $a0 = zeroes(3,2);
 # $a0->doflow();
 
- print $a0;
+ note $a0;
```
mohawk2: `note` doesn't seem to add anything here.

zmughal:
I believe I have made most of the changes you suggested. I kept some as they are for now since I'm going to do another pass over the test suite with Perl::Critic after the split. I'll squash the last couple commits soon.

zmughal:
Looks like this can be merged now. What do you all think?

zmughal:
TODO:
- [ ] diff the test outputs wrt `master` to make sure no messages were lost
- [ ] run Perl::Critic on the tests
- [ ] rerun checks on what was listed in the top

d-lamb:
Looks good to me.

zmughal:
I'll finish the last couple TODOs this weekend just to make sure everything checks out and then do the long-awaited merge!

zmughal:
After doing a line-by-line comparison of the output, this is now ready to merge.

PR #111, opened 2015-05-30
Fix version build bug

mohawk2:


zmughal:
Travis-CI looks happy and this looks reasonable to me. This should prevent having to bump the version in multiple places.

mohawk2:
In fact, it's the bare minimum fix. Most of the other core modules are unversioned. That should probably change in the fullness of time. Anyway, I'm merging.

PR #112, opened 2015-05-30
Complete purge of PDL::Graphics::PLplot

mohawk2:


TODO
```diff
@@ -1,12 +1,6 @@
 What follows is that task list and description of
 the work planned for the future  PDL-2.x releases.
 
-The main goal we are working towards is to have a
```
jberger: why is this paragraph to be removed and not just the mention of PLplot?

PR #113, opened 2015-05-30
Pdl core only build - DO NOT MERGE

mohawk2:
This is for Travis purposes (and for using to release code). The master branch should not have this until the `kitchen-sink` directory is empty. Instead, this will be rebased along as necessary.


mohawk2:
I am declaring victory on this, since all of the ones in "kitchen-sink" have indeed been either incorporated because considered "core" enough, or now independent CPAN distros.

PR #121, opened 2015-06-14
New changelog format

zmughal:
This is a reopening of PR #78 in order to address #67 which makes the `Changes` format follow a standard format.


zmughal:
This PR will be obsoleted by #123 when that is merged.

PR #122, opened 2015-06-14
Add PDL::Doc::add_module to pdlpp_postamble

mohawk2:
This means eg PDL::IO::GD as an independent module will get pdldoc-ed just like modules in the monolith distro.


zmughal:
For dists like [PDL-Stats](https://github.com/maggiexyz/PDL-Stats/blob/master/add_doc.pl) and [PDL-Parallel-threads](https://github.com/run4flat/PDL-Parallel-threads/blob/master/Build.PL#L13) which use `scantree`, how will they be affected?

zmughal:
Do we have a test to ensure that the docs have been added?

PR #123, opened 2015-06-14
Fix for force-push post-2.012

zmughal:
# ~~DO NOT MERGE~~
> ~~Waiting until PDL 2.012 is known to have indexed properly~~

This drops the commit that added 2.011_01, since that was not released and we went straight to 2.012 from 2.011.

Edit: The PDL 2.012 release is working.


zmughal:
This is ready to go!

mohawk2:
Looks good to me.

mohawk2:
However, let's only merge if we have buy-in from @devel-chm.

Issue #124, opened 2015-06-17
SF#390 scalar PDL with badvalue always compares BAD with perl scalars


zmughal:
From <http://sourceforge.net/p/pdl/bugs/390/>

> This problem was reported by as user summarized in this message:
> https://sourceforge.net/p/pdl/mailman/message/34211884/
> as shown in this sample program:
>
> ```perl
> use PDL;
> use PDL::NiceSlice;
>
> my $x = pdl(1, 2, 3, 0);
> $x->badflag(1);
> $x->badvalue(0);
>
> print "x = $x\n";
>
> my ($m, $s) = stats($x);
>
> print "m = $m, s = $s\n";
> print "s greater than zero\n" if $s > 0;
> print "s less than zero\n" if $s < 0;
> print "s equals zero\n" if $s == 0;
> ```
>
> In fact, other comparisons of a scalar PDL value with the badflag set show the same error:
>
> ```perl
> pdl> $s = pdl(0)
>
> pdl> $s->badflag(1)
>
> pdl> p $s
> 0
>
> pdl> p $s>3
> 0
>
> pdl> p $s->badvalue(0)
> 0
>
> pdl> p $s>3
> BAD
>
> pdl> $s .= 2
>
> pdl> p $s
> 2
>
> pdl> p $s>3
> BAD
>
> pdl> p $s->badvalue(35)
> 35
>
> pdl> p $s>3
> 0
>
> pdl> p $s->badvalue(2)
> 2
>
> pdl> p $s>3
> BAD
> ```
>
> Given the weird and not yet understood behavior, I have marked this as highest priority.

zmughal:
I have put together some tests on the sf390 branch:
<https://github.com/PDLPorters/pdl/pull/128>.

I think I may have a direct fix for the problem regarding
stats(). The following test shows what I believe is a fundamental issue
with the logic found in stats()/statsover() [^1]:

	subtest "stats() should not set the badflag for output" => sub {
		my $p = pdl [1, 2, 3];
		$p->badflag(1);
		$p->badvalue(2);

		diag "\$p = $p";
		is( "$p", "[1 BAD 3]");

		my $m = stats($p);

		is( "$m", "2", "Mean of [1 3] is 2" );
	};

The PDL $p in the above stringifies to "[1 BAD 3]" since 2 is now a
badvalue. Now if I calculate the mean of $p, it should ignore the
badvalue and calculate ( 1 + 3 ) / 2 == 2.

[...]

PR #128, opened 2015-06-21
test comparisons between scalars and 0-dim PDLs with badflag == 1

zmughal:
This is to elucidate the issues with stat() and per-PDL badvalues
brought up by Marek Gierliński in SF#390
<http://sourceforge.net/p/pdl/bugs/390/>,
<https://github.com/PDLPorters/pdl/issues/124>.

### DIFF
--- a/Basic/Ops/ops.pd
+++ b/Basic/Ops/ops.pd
@@ -89,7 +89,9 @@ EOH
     # handle exceptions
     my $badcode = '$ISBAD(a()) || $ISBAD(b())';
     if ( exists $extra{Exception} ) {
-	delete $exists{Exception};
+	# NOTE This option is unused ($badcode is not set).
+	#      See also `ufunc()`.
+	delete $extra{Exception};
     }
 
     pp_def($name,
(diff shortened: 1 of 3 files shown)


zmughal:
Thread on pdl-devel <http://thread.gmane.org/gmane.comp.lang.perl.pdl.general/8729/focus=6584>

zmughal:
Plan for further work is discussed at <http://irclog.perlgeek.de/pdl/2015-07-18#i_10918794>:

  - create a warning for when a comparison op has *_badval ∈ {0, 1} ⇒ merge to master & do a full release
  - change the output type for comparisons to byte ⇒ test on devel branch, see what breaks, & do a dev release

zmughal:
Plan for further work is discussed at <http://irclog.perlgeek.de/pdl/2015-07-18#i_10918794>:

  - create a warning for when a comparison op has *_badval ∈ {0, 1} ⇒ merge to master & do a full release
  - change the output type for comparisons to byte ⇒ test on devel branch, see what breaks, & do a dev release

mohawk2:
Why not just unconditionally `delete $extra{Exception};` (lines 92-6 of ops.pd)? Similarly, lines 248-54.

I'm also keen to squash together / eliminate failing commits, to ease future bisecting.

The bit of primitive.pd you changed looks like it's crying out for a bit of data-driven action.

zmughal:
I squashed and reordered the commits.

I want to hold off on the two things you mentioned:
  - turning the badvalue propagation section in `statsover()` into a loop that applies the same to each return value: this would not give that much reduction in complexity. Not enough to justify having a a `for`/`map` to generate code.
  - removing the `Exception` key in `ops.pd`: the `Exception` is actually a placeholder for more sane ops...it is just that it was never implemented properly. There's a bug there and I'd rather leave it in for now as a `grep`pable reminder.

PR #138, opened 2015-08-17
Set graticule() to scalar context in Cartography demo.

### LINKED ISSUE #137: SF#394 demo cartography fails
From <https://sourceforge.net/p/pdl/bugs/394/>

> Testing the demos in the latest PDL-2.013 release, the cartography demo resulted in a crash.
> Investigating, the problem is in the Demos/Cartography_demo.pm file where the map and
> graticule are generated/loaded:
> 
> 
>     ### Get the vector coastline map (and a lon/lat grid), and load the Earth
>     ### RGB daytime image -- both of these are built-in to the module. The
>     ### coastline map is a set of (X,Y,Pen) vectors.
>     $coast = earth_coast() -> glue( 1, graticule(15,1) );
>     print "Coastline data are a collection of vectors:  ",
>        join("x",$coast->dims),"\n";
> 
> Yielding the error:
> 
>       OOPS!!! Something went wrong, please make a bug report!: PDL: PDL::Primitive::append(a,b,c):
>       Parameter 'b':  Mismatched implicit thread dimension 0: size 3 vs. 2
>       There are 3 PDLs in the expression; 1 thread dim....
> 
> The problem arises because graticule() returns 2 pdls in list context but
> one of shape [3,N] for scalar context.  Scalar context must be forced for
> the thread dimensions to match:
> 
>      $coast = earth_coast() -> glue( 1, scalar(graticule(15,1)) );

d-lamb:
Just a simple 7-byte change, should be good to go. This fully resolves #137

### DIFF
--- a/Demos/Cartography_demo.pm
+++ b/Demos/Cartography_demo.pm
@@ -59,7 +59,7 @@ act q|
   ### Get the vector coastline map (and a lon/lat grid), and load the Earth
   ### RGB daytime image -- both of these are built-in to the module. The
   ### coastline map is a set of (X,Y,Pen) vectors.
-    $coast = earth_coast() -> glue( 1, graticule(15,1) );
+    $coast = earth_coast() -> glue( 1, scalar graticule(15,1) );
     print "Coastline data are a collection of vectors:  ",
              join("x",$coast->dims),"\n";
 


zmughal:
Looks good to me!

We don't have tests for the demos yet, so any CI checks don't really mean much. This is simple enough to where don't have to worry.

**Merge**: *Aye*

@devel-chm, @mohawk2, this looks ready to go into `master`, yes? Merge and add to the v2.013_01 milestone?

mohawk2:
Make it so! And dev-release it!

PR #144, opened 2015-08-26
Travis-CI (smoke longlong-double-fix)

kmx:
The purpose of this pull request is to smoke current longlong-double-fix branch. It corresponds to commit https://github.com/PDLPorters/pdl/commit/e3304520cca34f91e308c7175e2a250b2c682256 (rebased on master and squashed into 1 commit)

PR #145, opened 2015-08-28
longlong double fix

zmughal:
This branch is for improving the 64-bit PDL support. This addresses issue #76, <https://github.com/PDLPorters/pdl/issues/37>, SF#343.

All changes to this branch will automatically be tested. This makes using the squashed commits in #144 unnecessary.

CC: @devel-chm, @kmx


devel-chm:
Apparently, AUTOMATED_TESTING is not being set by Travis-CI

PR #149, opened 2015-09-16
add a warning if the optional PGPLOT dependency is not available

zmughal:
This warns if PGPLOT is missing when is not installed because optional
dependencies can be skipped at installation. This gives users a clear
notification of why their code will not load.

Mithaldu++ for bringing up the issue on the #pdl IRC
http://irclog.perlgeek.de/pdl/2015-09-15#i_11219756.

CC: @wchristian 

Branch note: This branch is a WIP. It still needs tests and Travis-CI support. Anyone else is welcome to move it forward.


mohawk2:
I've updated this to still die, else `t/pgplot.t` fails (it relies on the `use` die-ing if no PGPLOT). However, there has been enough time to examine this, and merely adding more info to an exception should not be controversial. Merging.

PR #153, opened 2015-09-27
Mark CI testing using CI_TESTING environment variable

zmughal:
This is used to skip `t/bigmem.t` so that it does not die due to an OOM condition.


zmughal:
@devel-chm: this branch is passing and makes sure `t/bigmem.t` is skipped on both Travis-CI and Appveyor.

PR #176, opened 2016-02-22
Some changes to PDL::Fit::LM

d-lamb:
Mostly it is just documentation changes.  However, the new t/lm.t contains sample data copied from Gnuplot, which could potentially raise a license issue.  Check that file please.


mohawk2:
Since one option for Perl is to use under the GPL, I don't foresee a problem. If anyone ever complains, then we can revisit.

PR #178, opened 2016-03-16
Require ccNcompt return type to be >= long.

d-lamb:
Add 2 tests to check for this for byte image inputs.


devel-chm:
Looks good to me, Derek.

PR #186, opened 2016-06-07
add 64-bit support to PDL::IO::FITS

drzowie:
Fixes bug #421 (no 64 bit support for wfits/rfits)

### DIFF
--- a/IO/FITS/FITS.pm
+++ b/IO/FITS/FITS.pm
@@ -592,6 +592,7 @@ our $type_table = {
     8=>$PDL_B,
     16=>$PDL_S,
     32=>$PDL_L,
+    64=>$PDL_LL,
     -32=>$PDL_F,
     -64=>$PDL_D
 };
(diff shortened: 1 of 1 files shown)


drzowie:
OK, so I admit to being a bit slow here.  Does this mean the pull was accepted?  Or that it vanished into hyperspace?  

Checking out branch master, I notice that the mods seem to have vanished…

zmughal:
I see in the IRC log that you pushed the changes for 10494df4caa9e47ec64770a8eaa0c1674bb9b331 twice and they got deleted each time.

This is because, right now, the Sourceforge Git repository is considered the authoritative source. When a change appears on GitHub that isn't on Sourceforge, it gets rolled back through @wchristian's synchronisation cron job.

The approach I take is to make my `origin` remote push to both Sourceforge and GitHub (replace the sf.net username with your own):

```shell
git remote set-url --push origin git@github.com:PDLPorters/pdl.git
git remote set-url --add --push origin ssh://zsmughal@git.code.sf.net/p/pdl/code
```

Now when I do a push, it goes to both places with one command.

drzowie:
Sweet!  Thanks for the pro tip, Zaki.

(Mobile)

PR #189, opened 2016-06-27
Sf379 qsort crash

d-lamb:


d-lamb:
Probably.  I didn't realize until after I pushed the first commit that that branch was 19 commits behind master--I rebased master, then had to manually resolve the second commit.  Turned into a mess.

mohawk2:
Please do rebase. I had a total nightmare picking through Chris's LLDF commits to make sure I'd got them all, because of all the multi-way merges. It's worth it!

d-lamb:
OK, I have the two "content" commits cherry-picked into a new clean branch, which will be rebased onto master without the double-commit and the merge commit.  thanks for the heads-up that it got really ugly!

PR #193, opened 2016-08-22
Add datatype invalid

d-lamb:
See comments for commit f06ae90.


zmughal:
This needs to be rebased, right?

d-lamb:
yeah, it turned into a mess.  I think when I started that branch I accidentally based it off another one that isn't master.  There's really only one commit in there (f06ae90) that is relevant for this branch.  Will try again this weekend.

PR #195, opened 2016-08-29
Convert a bunch of tests to Test::More, etc.

d-lamb:


d-lamb:
@zmughal can you kick off the Travis CI builds again?  The fact that the first one took 13+ minutes makes me suspicious.  Or is there a way to get into the instances of those failing tests and manually run the tests to get some additional diagnostics?

zmughal:
I've restarted the failing builds again. You can do the same if you log in to Travis CI with your GitHub account.

From the logs, I see that one of the builds had a failure in `t/image2d.t` (not all the tests ran?) and `t/lvalue.t` (seemed to exit early?). We'll see if that was a fluke after the builds finish running.

By the way, I had some similar changes in #109 which hasn't been merged in yet. I can rebase it off master. It'll be a bit tricky, but that hasn't stopped me before! :-)

mohawk2:
Can I suggest we get both sets of changes rebased, deconflicted and merged? It hurts me to see such good work sitting there on a branch :-)

d-lamb:
I'm able to reproduce the `t/image2d.t` failure on my machine building with no bad value support.  Will investigate.  I have no idea why `t/lut.t` would be exiting early.

I completely missed #109 when I started doing this, sorry.  Let's get these both into master!

d-lamb:
thanks for the pointers on how to kick off the builds again.  I saw the "login with your Github account" button, but then when Travis wanted my personal information and my repositories and my first born I backed away quickly.

d-lamb:
For `t/lvalue.t`, when running under Devel::Cover on my machine, in the BEGIN block the $PERLDB variable is 260, which causes it to drop into the `else` part of that block, which I didn't fix up properly.  For reference, 260 = 0x104, which means: `0x100: Provide informative "file" names for evals based on the place they were compiled` and `0x004: Switch off optimizations`.

zmughal:
This looks good to go.

**Merge**: *Aye*.

Also, I just figured out the reason why I hadn't merged <https://github.com/PDLPorters/pdl/pull/109>. I had merged it in pdla as <https://github.com/PDLPorters/pdla/pull/7> instead.

PR #214, opened 2018-04-09
Change from Sourceforge to GitHub

mohawk2:
Building on the excellent work of Derek, and as discussed on the `pdl-devel` mailing list, I believe this covers all of the changes required to transition to using GitHub for PDL development going forward.

I reproduce here my proposed list of needed actions:
- [x] ensure all the people who should have push access are on GH and are in the PDLPorters organisation
- [x] ensure all SF issues are present as GH issues
- [x] announce say a week in advance we will be switching the repo location
- [ ] CODE FREEZE as releases are going to be happening
- [ ] at the same time, dev-release new version with the updated: repo location, bugtracker location
- [ ] after that week, proper-release that version
- [ ] quick final check that GH and SF repos are in sync
- [ ] turn off Christian’s script – no more error emails!
- [ ] sweep up stragglers who should be in PDLPorters but aren't

I would suggest the intent is already clear, so I have marked some of these done. If everyone is happy with these changes, I would suggest the next stage is merge (which I will do), then Chris can announce code freeze, dev-release, then when he is happy, main-release.

### DIFF
--- a/Basic/Pod/FAQ.pod
+++ b/Basic/Pod/FAQ.pod
@@ -390,8 +390,8 @@ and pdl-devel was called 'pdl-porters'.
 |------------|-------------------------------------------------------|
 |1996 - 2004 | http://www.xray.mpe.mpg.de/mailing-lists/perldl/      |
 |1997 - 2004 | http://www.xray.mpe.mpg.de/mailing-lists/pdl-porters/ |
-|2005 - 2015 | perldl@jach lost???                                   |
-|2005 - 2015 | pdl-porters@jach lost???                              |
+|2005 - 2015 | http://perldl.jach.hawaii.narkive.com/                |
+|2005 - 2015 | http://pdl-porters.jach.hawaii.narkive.com/           |
 |2015 -      | https://sourceforge.net/p/pdl/mailman/pdl-general/    |
 |2015 -      | https://sourceforge.net/p/pdl/mailman/pdl-devel/      |
 |--------------------------------------------------------------------|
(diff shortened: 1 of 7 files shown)


mohawk2:
@karlglazebrook It's the one running every 5(?) minutes that mirrors from the SF repo to the GH one, enabling us to do all the CI and Coveralls magic! But it's having lots of errors due to SF's unreliability so it's time to retire it.

Basic/Pod/FAQ.pod
```diff
@@ -367,30 +367,34 @@ pages).
...
+and pdl-devel was called 'pdl-porters'.
+
+|Time Period | URL                                                   |
+|------------|-------------------------------------------------------|
+|1996 - 2004 | http://www.xray.mpe.mpg.de/mailing-lists/perldl/      |
+|1997 - 2004 | http://www.xray.mpe.mpg.de/mailing-lists/pdl-porters/ |
+|2005 - 2015 | http://perldl.jach.hawaii.narkive.com/                |
+|2005 - 2015 | http://pdl-porters.jach.hawaii.narkive.com/           |
+|2015 -      | https://sourceforge.net/p/pdl/mailman/pdl-general/    |
+|2015 -      | https://sourceforge.net/p/pdl/mailman/pdl-devel/      |
```
zmughal: Very nice. This would be good to have on the website as well!

mohawk2:
@devel-chm Agreed. That was my plan - as many stages as possible!

zmughal:
Documentation changes look good! :+1:

Basic/Pod/FAQ.pod
devel-chm: On my wishlist is a better mailing list archive with searching.  I have the files from the "early days" but don't have a solution for the mail database/searching.

PR #233, opened 2018-04-19
Devrel2 fixups

mohawk2:
This temporarily reverts Jerry's patch for MatrixOps. When this is merged to master, I will make a PR to add it back in. I can't right now since SF's repo is giving

> `fatal: remote error: access denied or repository not exported: /p/pdl/code`

yet again. @d-lamb I saw your excellent work on a second `sf_to_gh`, can you just push that straight to master? Then @devel-chm could you merge this as well, then make a second dev-release?

### DIFF
--- a/Known_problems
+++ b/Known_problems
@@ -69,11 +69,7 @@ The following issues have been reported with this version of PDL:
   again.  The problem has been reported as a Term::ReadLine::Perl bug.
 
 
-- The following bugs are outstanding at time of the PDL-2.018_01 release:
-
-  ===============================================
-  TODO: need refs to github bugs here, not sf.net
-  ===============================================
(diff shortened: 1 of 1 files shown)


d-lamb:
I merged that sf_to_gh into master again.

mohawk2:
Since the build worked on Derek's non-code changes, I have to assume the breakage we saw (in ufunc) is not connected to Jerry's change to MatrixOps. I therefore see no reason to revert that now.

@devel-chm Do you see any value in the remaining change on here, which updates your change to `Known_problems`, or shall I just close this? Also, what do you think about now releasing 2.019?

devel-chm:
Let's do one more developers release (I can make tonight) for final review and checkout.  Then we can have a PDL 2.019 release this weekend.

Issue #222, opened 2018-04-09
SF#436 pdldoc dims shows only PDL::IO::HDF5::Dataset and not PDL::Core dims


mohawk2:
@devel-chm writes:

    pdldoc dims
    Module PDL::IO::HDF5::Dataset
    dims
    Get the dims for a HDF5 dataset. For example, a 3 x 4 array would return
    a perl array (3,4);

    Usage:

     @pdl = $dataset->dims;    # Get an array of dims.

    Docs from
    /local64/lib/perl5/cygwin-thread-multi/PDL/IO/HDF5/Dataset.pm

but if I look at the PDL distribution result I get:

    perl -Mblib pdldoc dims
    Module PDL::Core
    dims
    Return piddle dimensions as a perl list

     @dims = $piddle->dims;  @dims = dims($piddle);

     pdl> p @tmp = dims zeroes 10,3,22
     10 3 22

    See also shape which returns a piddle instead.

    Docs from /pdl/git/pdl-code/blib/lib/PDL/Core.pm

so it appears that the install of the PDL docs from the PDL::IO::HDF5 module has overwritten the separate dims() from PDL::Core. Ideally, both should be present and listed in the default pdldoc output.

This is for PDL-2.018 with PDL::IO::HDF5 0.73.

https://sourceforge.net/p/pdl/bugs/436/

mohawk2:
As noted, #234 fixes. Closing.

PR #234, opened 2018-04-26
Fix documentation system

d-lamb:
If this branch were merged into master, nothing would break and usability would be improved.  As is, this accomplishes:
- functions whose name duplicates the name of a function in a different module now show up in the documentation database;
- the 'apropos', 'usage', 'sig', and 'badinfo' documentation helper functions now work with the documentation database's different naming convention

What remains is to account for functions that share documentation.  For example, mzeroes, mones, & msequence in PDL::Matrix all have an entry in the database, but do not appear to have any actual documentation that makes it in.  So doing 'help mzeroes' results in a blank entry.  One must be careful that the change that fixes this does not result in extraneous info in the other cases: using 'help inner' as a comparison is advised.

I believe that properly changing the regexp in [getfuncdocs in Doc.pm](https://github.com/PDLPorters/pdl/blob/5b9846e754ac1241e05f8fb2a6cdd56d91d2e1f6/Doc/Doc.pm#L814l) would solve that remaining issue, but I have not figured out the correct incantation.

If merged, #222 and #28 should be able to be closed.

### DIFF
--- a/Doc/Doc.pm
+++ b/Doc/Doc.pm
@@ -640,7 +640,7 @@ sub checkregex {
   return "(?i)$regex" unless $regex =~ /^m[^a-z,A-Z,0-9]/;
   my $sep = substr($regex,1,1);
   substr($regex,0,2) = '';
-  $sep = '(?<!\\\\)\\'.$sep; # Avoid '\' before the separator 
+  $sep = '(?<!\\\\)\\'.$sep; # Avoid '\' before the separator
 
   my ($pattern,$mod) = split($sep,$regex,2);
   barf "unknown regex modifiers '$mod'" if $mod && $mod !~ /[imsx]+/;
@@ -891,7 +891,7 @@ sub add_module {
 	}
 	return if($hit);
     }
-    
+
     die "Unable to find a .pm or .pod file in \@INC for module $module\n";
 }
 
--- a/Doc/Doc.pm
+++ b/Doc/Doc.pm
@@ -165,7 +165,7 @@ An implementation of online docs for PDL.
 =head1 Using PDL documentation
 
 PDL::Doc's main use is in the "help" (synonym "?") and "apropos"
-(synonym "??") commands in the perldl shell.  PDL:Doc provides the
+(synonym "??") commands in the perldl shell.  PDL::Doc provides the
 infrastrucure to index and access PDL's documentation through these
 commands.  There is also an API for direct access to the documentation 
 database (see below).
--- a/Doc/Doc.pm
+++ b/Doc/Doc.pm
@@ -506,7 +506,7 @@ sub savedb {
         $val->{File} = File::Spec->abs2rel($fi, dirname($this->{Outfile})) ;
       }
       delete $val->{Dbfile}; # no need to store Dbfile
-      my $txt = "$name".chr(0)."$module".chr(0).join(chr(0),%$val);
+      my $txt = join(chr(0),$name,$module,%$val);
       print OUT pack("S",length($txt)).$txt;
     }
   }
@@ -704,7 +704,7 @@ sub scan {
     if (defined($val->{Module})){
 	$hash->{$key}->{$val->{Module}} = $val;
     } else {
-	print STDERR "no Module for $key in $file2\n";
+	warn "no Module for $key in $file2\n";
     }
 }
 
--- a/Doc/Doc.pm
+++ b/Doc/Doc.pm
@@ -823,7 +823,7 @@ sub getfuncdocs {
 #  $parser->select("\\(METHODS\\|OPERATORS\\|CONSTRUCTORS\\|FUNCTIONS\\|METHODS\\)/$func(\\(.*\\)*\\s*");
   foreach my $foo(qw/FUNCTIONS OPERATORS CONSTRUCTORS METHODS/) {
       seek $in,0,0;
-      $parser->select("$foo/$func(\\(.*\\))*\\s*");
+      $parser->select("$foo/(.*,\\s+)*$func(\\(.*\\))*(\\s*|,\\s+.*)");
       $parser->parse_from_filehandle($in,$out);
   }
 }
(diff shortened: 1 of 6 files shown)


mohawk2:
Please can we do one thing at a time? First switch over the repo to GitHub, then other things. I would be very supportive of the next release being only this. (You may recognise this concept as being my hope we can "release early, release often" :-) - things that would make this much easier would be tooling to make @devel-chm able to release a version super-easily)

mohawk2:
@d-lamb @devel-chm @zmughal @drzowie Is this ready for primetime? :-)

d-lamb:
oof, it's been awhile since I've looked at it!  I will try to find some time in the next week or so.  My sense is that NO, it's not ready for primetime, for the reasons stated in the original post.

d-lamb:
THIS, now, I am happy with!  Considering the depth of the changes, and that most of the functionality is interactive, it needs some manual testing.

I have NOT tried with pdl2, but it looks like that shell pulls in PDL::Doc::Perldl so hopefully the fixes flow to that shell too.

Doc/Doc.pm
```diff
@@ -496,16 +497,18 @@ sub savedb {
...
+  while (my ($name,$mods_hash) = each %$hash) {
+    next if 0 == scalar(%$mods_hash);
+    while (my ($module,$val) = each %$mods_hash){
+      my $fi = $val->{File};
+      if (File::Spec->file_name_is_absolute($fi) && -f $fi) {
+        #store paths to *.pm files relative to pdldoc.db
+        $val->{File} = File::Spec->abs2rel($fi, dirname($this->{Outfile})) ;
+      }
+      delete $val->{Dbfile}; # no need to store Dbfile
+      my $txt = "$name".chr(0)."$module".chr(0).join(chr(0),%$val);
```
mohawk2: Maybe just `my $txt = join(chr(0), $name=>$module, %$val);`?

Doc/Doc.pm
```diff
@@ -683,12 +696,17 @@ sub scan {
...
+
     $n++;
 
     $val->{File} = $file2;
-    $hash->{$key} = $val
+    #set up the 3-layer hash/database structure: $hash->{funcname}->{PDL::SomeModule} = $val
+    if (defined($val->{Module})){
+	$hash->{$key}->{$val->{Module}} = $val;
+    } else {
+	print STDERR "no Module for $key in $file2\n";
```
mohawk2: Not a `warn`?

Doc/Doc/Perldl.pm
```diff
@@ -77,7 +86,7 @@ sub screen_width() {
     return $ENV{COLUMNS}
        || (($ENV{TERMCAP} =~ /co#(\d+)/) and $1)
        || ($^O ne 'MSWin32' and $^O ne 'dos' and 
-	   (`stty -a 2>/dev/null` =~ /columns\s*=?\s*(\d+)/) and $1)
+	  ((`stty -a 2>/dev/null` =~ /columns\s*=?\s*(\d+)/) or (`stty -a 2>/dev/null` =~ /(\d+)\s+columns/)) and $1)
```
mohawk2: Is this not encapsulated in a CPAN module?

Doc/Doc.pm
d-lamb: Even better, `my $txt = join(chr(0),$name,$module,%$val);`.

Doc/Doc.pm
d-lamb: good catch.  mostly that was for debugging during development but it's good to keep in there.

Doc/Doc/Perldl.pm
d-lamb: Yes.  [perlfaq8](https://perldoc.perl.org/5.30.0/perlfaq8.html#How-do-I-get-the-screen-size%3f) suggests using [GetTerminalSize](https://metacpan.org/source/JSTOWE/TermReadKey-2.38/ReadKey.pm.PL#L399) in [Term::ReadKey](https://metacpan.org/pod/Term::ReadKey), and we could add that as a CPAN dependency since Term::ReadKey is not a Perl core module.

FWIW that function, after some handling of more esoteric systems (which may include Windows 😕) , seems to look for either
- $ENV{COLUMNS}
- the `resize` command (comes with X11 or other window managers), or
- `stty size` (which may be a cleaner than `stty -a`)

If foisting the cross-platform compatibility issues onto somebody else is worth adding a dependency, then that's fine with me.

mohawk2: If we need to get this information, best not to try to do it ourselves - PDL is about whizzy data, not about terminal-handling :-)

Derek, do you feel like also switching this code to use Term::ReadKey? I'm on the verge of full-releasing PDL, in a few days, and would like to then be able to merge this and push out an immediate dev-release with it, for people to try.

d-lamb: Sure, that's fine.  I cooked up a Term::ReadKey -optional version but then since it was so easy to install from CPAN I just got rid of all of our old code.  For my own reference, that half-and-half version looked like:

~~~~perl
sub screen_width() {
    local $@;
    eval { require Term::ReadKey; };
    unless ( $@ ){
	return ( ( Term::ReadKey::GetTerminalSize(\*STDOUT) )[0] );
    } else {
    return $ENV{COLUMNS}
       || (($ENV{TERMCAP} =~ /co#(\d+)/) and $1)
       || ($^O ne 'MSWin32' and $^O ne 'dos' and
	  ((`stty -a 2>/dev/null` =~ /columns\s*=?\s*(\d+)/) or (`stty -a 2>/dev/null` =~ /(\d+)\s+columns/)) and $1)
       || 72;
    }
}
~~~~
But now it's just
~~~~perl
use Term::ReadKey
sub screen_width() {
    return ( ( GetTerminalSize(\*STDOUT) )[0] );
}
~~~~
My only concern with this approach is that somebody who doesn't have Term::ReadKey already installed will have a broken perldl shell install if they do an upgrade and don't see the one-line warning about not having the new prerequisite.  Should Makefile.PL die when those prerequisites aren't satisfied?

d-lamb:
@mohawk2, I am sorry, I misread your comment about the order of 2.020 and merging this in.  If you want to revert the merge to keep 2.020 clean, I'm OK with that.

mohawk2:
I have moved the merge to after 2.020, which I have just released, and released 2.020_01 with the merge. Please be more careful.

PR #235, opened 2018-05-09
Honor LDFLAGS when building pdl.

0-wiz-0:
This is needed e.g. when building with RELRO support.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -586,7 +586,7 @@ EOPS
 # support the `$<` variable in explicit rules
 $text .= <<EOT if $^O !~ /MSWin/;
 pdl$Config::Config{exe_ext} : pdl.c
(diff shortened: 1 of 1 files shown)


0-wiz-0:
I've used it for -Wl,-z options and it worked fine where it was.

mohawk2:
Looks good to me. @devel-chm want to merge?

zmughal:
Since the above checks do not show the **passing** AppVeyor build (probably a notification hiccup), here it is: <https://ci.appveyor.com/project/zmughal/pdl/build/1.0.627>.

zmughal:
This seems reasonable.

Merge: **Aye**.

mohawk2:
@0-wiz-0 Thanks very much for this! Going forwards, it will be better if you do it on a local branch that's not `master`, which makes rebasing a bit less complicated should it be needed :-)

PR #239, opened 2018-05-15
Fix spelling errors.

sebastic:
The lintian QA tool reported a couple of spelling errors for the PDL 2.019 Debian package build:

 * severly  -> severely
 * arbitary -> arbitrary


mohawk2:
Thanks for these!

Issue #238, opened 2018-05-15
PDL 2.019 fits.t fails on Debian unstable (tilde expansion didn't fail)


sebastic:
The pdl Debian package failed to build after updating it to 2.019 due to two `fits.t` tests failing:
```
#   Failed test 'wfits tilde expansion didn't fail'
#   at t/fits.t line 322.
# died: Unable to create FITS file >/nonexistent/PDL-IO-FITS-test_27901.fits
#  at /build/pdl-2.019/blib/lib/PDL/IO/FITS.pm line 1913.
#       PDL::wfits() called at t/fits.t line 322
#       Test::Exception::lives_ok() called at t/fits.t line 322

#   Failed test 'rfits tilde expansion didn't fail'
#   at t/fits.t line 323.
# died: FITS file /nonexistent/PDL-IO-FITS-test_27901.fits not found at /build/pdl-2.019/blib/lib/PDL/IO/FITS.pm line 355.
#       PDL::rfits() called at /build/pdl-2.019/blib/lib/PDL/IO/FITS.pm line 571
#       PDL::IO::FITS::rfits() called at t/fits.t line 323
#       Test::Exception::lives_ok() called at t/fits.t line 323
Could not delete /nonexistent/PDL-IO-FITS-test_27901.fits: No such file or directory
# Looks like you failed 2 tests of 97.
...
not ok 96 - wfits tilde expansion didn't fail
not ok 97 - rfits tilde expansion didn't fail
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/97 subtests 
        (8 TODO tests unexpectedly succeeded)
```

Full buildlog: [pdl_2.019-1_amd64.txt](https://github.com/PDLPorters/pdl/files/2003621/pdl_2.019-1_amd64.txt)

While on most systems `~` will be writable, this should be tested explicitly, as the Debian package build environment uses a non-existent `$HOME`.

PR #240, opened 2018-05-15
Fix tilde expansion tests for environments with non-existent $HOME.

sebastic:
As reported in #238 the Debian package build of PDL 2.019 failed because the build environment uses a non-existent `$HOME`.

The failure was fixed by only reading & writing the test file if the directory is writable.

Fixes: #238

### DIFF
--- a/t/fits.t
+++ b/t/fits.t
@@ -3,6 +3,8 @@
 
 use strict;
 
+use File::Basename;
+
 use PDL::LiteF;
 
 use PDL::Core ':Internal'; # For howbig()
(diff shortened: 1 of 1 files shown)


mohawk2:
@sebastic I agree with the principle of this, but I think the specifics could be tweaked. Could you instead wrap the whole "check tilde expansion" block in a suitable `if`, with a comment? I'm not quite loving the idea of a `-w` check.

sebastic:
I'm not a fan of moving the number of tests to the `done_testing()` call at the end, but it works too.

@mohawk2, are you okay with the changes in 6ca3fd5?

mohawk2:
This looks fine. @sebastic could I ask you to get rid of the "number of tests" thing entirely? It's basically a historical thing, and `done_testing` by itself is just fine. Thanks for sticking with this, we're nearly there!

sebastic:
Done in 0a2f5b1

mohawk2:
Thanks for this! Fine work.

Issue #249, opened 2018-09-25
2.019 fails with perl since v5.27.1-63-g0599cd66b4 when compiled with -DEBUGGING=both


d-lamb:
...as reported by @andk at [here](https://sourceforge.net/p/pdl/bugs/445/)

This might be a good [example CPANtesters report](http://www.cpantesters.org/cpan/report/482292a6-837b-11e8-98a3-32f14d326c6b).

d-lamb:
As I commented on #250,  [this patch to Perl's pp_hot.c](https://rt.perl.org/Public/Bug/Display.html?id=131975) seems likely to be the change [described in perl5280delta.pod](https://perl5.git.perl.org/perl.git/blob/2dcd19f068f5fdcbcce36cc7a4ef94c53d3a47f3:/pod/perl5280delta.pod#l1977) that is triggering the failures.  The pp_hot.c patch was in response to [this patch to Perl's pp.h](https://perl5.git.perl.org/perl.git/blobdiff/23c687d9d091a545afb2b769447c17fba98ba87a..87058c31e9fa350bda8d797127c9c175d0b1a893:/pp.h) that updates a High Water Mark (hwm) for a calling function's stack when the XS EXTEND macro is called.

So far I have taken the following steps to diagnose this issue.

1. Built a debugging Perl-5.28.0 locally on my machine using Perlbrew.
2. Built PDL against that Perl, and verified that 'make test' does in fact fail.
3. Focused on PDL::Fit::Gaussian as that is the first test failure, and also a conceptually easy module (there are only 2 subroutines).
4. Commented out the pp_def for fitgauss1dr, to focus on fitgauss1d only.
5. Verified that calling fitgauss1d with the form:
`my ($xc1, $pk1, $fwhm1, $back1, $err1, $fit1) = fitgauss1d($x1, $g1);`
DOES cause the Panic message, but calling it with the (supposedly equivalent form):
`fitgauss1d($x1, $g1,$xc1=null,$pk1=null,$fwhm1=null,$back1=null,$err1=null,$fit1=null);`
DOES NOT cause the Panic message.  Calling it with all the temporaries=null also does not cause the Panic message.
6. Verified that the same fix causes the "max2d_ind" test in image2d.t to not fail, i.e, changing
`my @ans = $pa->max2d_ind();` to
[...]

PR #250, opened 2018-09-25
Add to .travis.yml a debugging blead perl.

d-lamb:
May help to diagnose #249.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ matrix:
 #    - perl: "5.16"
 #    - perl: "5.18"
     - perl: "blead"
+    - perl: "blead-dbg"
     # this is commented out because we test Perl v5.20 under all BADVAL configurations including the default below
     #- perl: "5.20"
     - perl: "5.20"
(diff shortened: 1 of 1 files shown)


d-lamb:
[This patch](https://rt.perl.org/Public/Bug/Display.html?id=131975) seems likely to be the change [described in perl5280delta.pod](https://perl5.git.perl.org/perl.git/blob/2dcd19f068f5fdcbcce36cc7a4ef94c53d3a47f3:/pod/perl5280delta.pod#l1977) that is triggering the failures.

mohawk2:
I think we need to update the 5.20 Perls in the Travis to 5.28, which both has this code change, and is also post-5.26 and will therefore catch any .-in-INC issues.

@d-lamb Any idea why this update to `.travis.yml` isn't failing? It *is* using a blead perl, yet the tests are passing?

mohawk2:
Ah, my bad - the `blead-dbg` *is* failing, but it's an "allowed failure". Switching from 5.20 to 5.28 should also fail, and show as a Travis failure.

mohawk2:
That seems to have nailed it! Fine work @d-lamb!

PR #253, opened 2018-12-05
Make functions in Types.pm visible to pdldoc.

d-lamb:
Because there was no '=head1 FUNCTIONS' the functions never got
indexed.  Also I added some other POD directives and examples.


mohawk2:
Looks good to me.

PR #254, opened 2018-12-05
Use Config{WITH_SLATEC} in slatec-polyfit-weight.t startup

d-lamb:
This makes the test planning consistent between the two slatec
test programs.


mohawk2:
Looks good to me. I wonder whether that check really needs to be a `BEGIN`, but never mind.

Issue #248, opened 2018-09-22
missing $GENERIC() in ufunc.pd b*over() BadCode


d-lamb:
@moocow-the-bovine [reported this](https://sourceforge.net/p/pdl/bugs/446/) on the old SF.net bug tracker:

> I use PDL/Basic/Ufunc/ufunc.pc code to generate "expected" results for tests of PDL::CCS sparse pseudo-piddles, and am getting test failures from my sparse borover() on armv6l-linux (rasberry pi), apparently due to bogus "expected" values returned by PDL::borover(); see [RT bug #126294](https://rt.cpan.org/Ticket/Display.html?id=126294).
> 
> **Minimum broken example:**
> ~~~
> use PDL;
> print pdl([10,0,-2])->setvaltobad(0)->borover();
> ~~~
> 
> **Expected output:**
> ~~~
> -2
> ~~~
> 
> **Apparent output on armv6l-linux:**
> ~~~
> 254
> ~~~
> 
> This only seems to occur in the presence of BAD values -- the same test without setvaltobad(0) succeeds.  It strikes me as suspicious that (254==(unsigned char)-2), and the associated [debian bug #902192](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902192) suggests that the problem is related to that fact that 'char' is unsigned on ARM.  Looking at the 'BadCode' section of the borover() and bandover() definitions in PDL/Basic/Ufunc/ufunc.pd, it seems that there's a literal 'char tmp' being used as an accumulator variable rather than the '$GENERIC()' datatype used in the corresponding 'Code' section (via $def=$over{$func}{def}). 
> 
> I myself have no direct access to an ARM machine for testing, but have a strong hunch that the 
> [attached patch](
> https://sourceforge.net/p/pdl/bugs/446/attachment/PDL-2.019_ufunc_bad_def.patch) might solve this issue.

moocow-the-bovine:
Thanks for moving this to the appropriate tracker, and apologies for reporting to the wrong place!

d-lamb:
I can confirm that if, instead of applying the attached patch, I change the accumulator in the BadCode section to be an 'unsigned char' (to mimic the behavior on the RasPi), I get the following results for borover and bandover:

~~~~perl
print pdl([10,0,-4])->borover()
-2
print pdl([-6,~0,-4])->bandover()
-8
pdl([10,0,-4])->setvaltobad(0)->borover()
254
[...]

PR #255, opened 2018-12-05
Fix borover and bandover when using bad values...

d-lamb:
...on systems with only unsigned chars.

This fixes GitHub issue #248. A rogue 'char tmp' instead of a
'$GENERIC() tmp' in the BadCode section of borover, bandover (and
also zcover, andover, and orover) caused problems on Raspberry Pi's.


mohawk2:
Looks good to me!

PR #257, opened 2019-01-11
Add logic to avoid segfault in some uses of PDL::Transform::map

d-lamb:
When using PDL::Transform::map with the 'g' method (pre-calculated
Gaussian sampling), if the inverse transform produced 'NaN' values
in the $idx variable, those NaNs got passed to _map_int as the
SV *map, and after PDL_xform_aux the variable 'tmp' would have
NaNs in it as well. Then in the input accumulation loop, the cp
variable was assigned tmp, which means dd and sum were now NaN.
In that case (sum > GAUSSIAN_MAXVAL) would be (NaN > GAUSSIAN_MAXVAL)
and that always evaluates to false, so essentially the calculation
of alpha would access illegal indices of the gaussian_lookup array.
This would cause a segmentation fault.

This segfault didn't happen for the 'h' method because the condition
if(beta > 0) was false, so the illegal access never happened.

I added a test (replicated for different resampling methods) that
triggers the segfault on the 'g' method without this commit's change
to transform.pd.

### DIFF
--- a/Lib/Transform/transform.pd
+++ b/Lib/Transform/transform.pd
@@ -843,7 +843,7 @@ pp_def('map',
        foo += ovec[k] * map->dimincs[k+1];
        printf(" %2d ",(int)(ovec[k]));
      }
-     printf("]; psize is %d; big is %d; blur is %8.2f; ret is %8g; map is [",psize,big, blur, ret);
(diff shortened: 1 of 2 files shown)


drzowie:
I pulled it into master.

PR #259, opened 2019-01-21
Range speedup2

d-lamb:
This has the same two commits that were on the original range-speedup branch, which had gone stale and was 400 commits behind master.  I messed up that branch with a rebase gone wrong (note: never start a 400-commit rebase at midnight), so just created this branch and cherry-picked the two commits of interest.

For reference, I ran the following code:
~~~~perl
use strict;
use warnings;
use PDL;
use PDL::NiceSlice;
use Benchmark;
my $im = sequence(1024,1024);
my $indx = random(2,3000)->mult(1023,0)->rint;
timethis (1E5,
	  sub {$im->range($indx,10,'t');}
    );
~~~~

Without these commits I get about 11000 iterations per second, and with the commits about twice that many:

~~~~pre
timethis 100000:  9 wallclock secs ( 9.10 usr +  0.01 sys =  9.11 CPU) @ 10976.95/s (n=100000)
timethis 100000:  5 wallclock secs ( 4.46 usr +  0.00 sys =  4.46 CPU) @ 22421.52/s (n=100000)
~~~~

### DIFF
--- a/Basic/Gen/PP.pm
+++ b/Basic/Gen/PP.pm
@@ -3486,11 +3486,17 @@ $PDL::PP::deftbl =
 
    # The RedoDimsSub rule is a bit weird since it takes in the RedoDims target
    # twice (directly and via RedoDims-PostComp). Can this be cleaned up?
-   #
+   #    [I don't know who put this in, or when -- but I don't understand it.  CED 13-April-2015]
    PDL::PP::Rule->new("RedoDims-PreComp", "RedoDims",
 		      sub { return $_[0] . ' $PRIV(__ddone) = 1;'; }),
    PDL::PP::Rule::MakeComp->new("RedoDims-PostComp",
 				["RedoDims-PreComp", "PrivNames", "PrivObjs"], "PRIV"),
+
+   # RedoDimsSub is supposed to allow you to use $SIZE as an lvalue, to resize things.  It hasn't
+   # worked since I can remember (at least since I started messing around with range).  The reason
+   # appears to be that the SIZE macro was using the redodims argument instead of its own zeroth
+   # argument.  Renaming gone wrong?  Anyway I've fixed it to use $_[0] instead of $redodims in the
+   # SIZE closure.   -- CED 13-April-2015
    PDL::PP::Rule->new("RedoDimsSub",
 		      ["RedoDims", "RedoDims-PostComp", "_DimObjs"],
 		      sub {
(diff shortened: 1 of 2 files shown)


Basic/Gen/PP.pm
```diff
@@ -3499,9 +3505,10 @@ $PDL::PP::deftbl =
 			my $dimobjs  = $_[2];
 
 			$result->[1]{"SIZE"} = sub {
-			  croak "can't get SIZE of undefined dimension (RedoDims=$redodims)."
-			    unless defined $dimobjs->{$redodims};
-			  return $dimobjs->{$redodims}->get_size();
+			    eval 'use PDL::IO::Dumper';
+			    croak "FOO can't get SIZE of undefined dimension (RedoDims=$redodims).\nredodims is $redodims\ndimobjs is ".sdump($dimobjs)."\n"
+			       unless defined $dimobjs->{$_[0]};  # This is the closure's $_[0], not the rule definition's $_[0]
```
mohawk2: Wouldn't this be better as something like:

```
			    # This is the closure's $_[0], not the rule definition's $_[0]
			    unless (defined $dimobjs->{$_[0]}) {
			      require PDL::IO::Dumper;
			      croak "FOO can't get SIZE of undefined dimension (RedoDims=$redodims).\nredodims is $redodims\ndimobjs is ".PDL::IO::Dumper::sdump($dimobjs)."\n";
			    }
```

This would avoid loading `PDL::IO::Dumper` unless there were an actual error, which would also make things a bit quicker.

Also, is it possible to add a test of this `SIZE` behaviour?

mohawk2:
@d-lamb Any thoughts on my comments above? I'd love to see this merged.

Basic/Gen/PP.pm
mohawk2:
```suggestion
			    # This is the closure's $_[0], not the rule definition's $_[0]
			    if (!defined $dimobjs->{$_[0]}) {
			      require PDL::IO::Dumper;
			      croak "FOO can't get SIZE of undefined dimension (RedoDims=$redodims).\nredodims is $redodims\ndimobjs is ".PDL::IO::Dumper::sdump($dimobjs)."\n";
			    }
```

mohawk2:
@d-lamb @devel-chm @zmughal @drzowie Is this ready for primetime? :-)

Basic/Gen/PP.pm
d-lamb: I think I looked at this and couldn't even figure out whether or where the SIZE behavior was actually used in PDL.  Closing and merging was the right thing to do here.

PR #260, opened 2019-02-28
Fix PDL::Lite

mohawk2:
Sadly, a22a11fb9b2e5527af94fbb3cc39745ef1215f55 actually caused `PDL::Lite` to correctly export functions `pdl` etc, that were no longer being imported. This broke, among other things, `Test::PDL`.

This PR fixes that, with a test to avoid in future.

### DIFF
--- a/Basic/Lite.pm
+++ b/Basic/Lite.pm
@@ -31,11 +31,9 @@ Access to other functions is by method syntax, viz:
 
 =cut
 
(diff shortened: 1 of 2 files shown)


mohawk2:
Merging after @karlglazebrook message expressing approval (thanks for taking a look): https://sourceforge.net/p/pdl/mailman/message/36638841/

PR #261, opened 2019-04-14
Coverage for XS as well

### LINKED ISSUE #92: Travis-CI: Make sure coverage includes XS files
See http://blogs.perl.org/users/nick_wellnhofer/2015/03/howto-xs-coverage-reports-on-coveralls.html.

mohawk2:
This is to fix #92.


mohawk2:
This code seems to work. It took some "clever" working around both `gcov` getting confused by XS stuff, and `Devel::Cover`' `gcov2perl` needing some hand-holding to get past problems with deep directory structure like PDL has.

The coverage levels shown are extremely low once the C/XS files etc are included, but this seems to actually reflect low test coverage.

mohawk2:
AppVeyor failure was because of DNS failure. Travis all passing. Merging.

PR #272, opened 2019-12-11
Nat c2

fantasma13:
Hi,
please checkout this branch for native complex support. It should hopefully apply against master without too many problems. 

I'm unsure of the changes to header file includes, in particular the proto.h caused issues, I'm not sure why it is required. It compiles fine and all tests pass on gcc 64 bit linux but I haven't tested other platforms. capn testers reports would be welcome for this.

The perldl.conf file may be modified, in particular testing the various bad value related options. 
There are a few substantial changes in the Core directory, Types.pm.PL, of course, and in particular the anyval_to_nayval function was moved to pdlcore.h.PL. Complex math is particular, so some spraying of GENERICTYPE was necessary throughout the ops, ufunc and math libs as well as some others.

For future support, I've added a few flags to data types (whether it's real or complex, integer or not and one for signed/unsigned. They are not (widely) used yet but could potentially facilitate writing fairly generic code. Feel free to scrap the idea if you don't like it.

Ingo

### DIFF
--- a/Basic/Bad/bad.pd
+++ b/Basic/Bad/bad.pd
@@ -534,7 +534,7 @@ EOC
     # - is there a better way of checking for the condition since
     #   the current one needs to be changed whenever the types are changed
     #
-    $set_code = "" if $usenan and ($type->ppsym eq "F" or $type->ppsym eq "D");
+    $set_code = "" if $usenan and ($type->ppsym eq "F" or $type->ppsym eq "D" or $type->ppsym eq "C" or $type->ppsym eq "G");
 
     $str .=
 "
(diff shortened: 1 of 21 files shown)


mohawk2:
I definitely like the idea of making the code more generic! Though it would need performance-testing.

In the mean time, please could you rebase against the current `master` branch? It looks like you've been using `git pull` and it's made merge commits, which makes the list of commits showing here very confusing! You'd go to your own `master` branch and bring that up to date.

I suspect I need to update the `DEVELOPMENT` file with best practice, since it's easy if you know how, but terrifying if you don't!

Finally, it looks like the build is failing on Windows. Obviously we need to get that working before we can think about merging.

Thanks for the work so far!

mohawk2:
I've squashed the many commits into one, since they weren't helpful to me in understanding the changes (good commits really help "tell a story"), and have rebased these changes against current `master`.

> and in particular the anyval_to_nayval function was moved to pdlcore.h.PL

I thought this was a good idea so have cherry-picked it to `master` (with you as the author).

PR #275, opened 2020-01-18
Fix pdlpp_postamble oneliner.

sebastic:
Building PDL::NetCDF failed with:
```
Updating PDL documentation database...
"/usr/bin/perl"  -e 'exit if \$\$ENV{DESTDIR}; use PDL::Doc; eval { PDL::Doc::add_module(q{$mod}); }' --
Scalar found where operator expected at -e line 1, near "$\$ENV"
        (Missing operator before $ENV?)
syntax error at -e line 1, near "$\$ENV"
BEGIN not safe after errors--compilation aborted at -e line 1.
```


mohawk2:
Good catch - thanks!

PR #276, opened 2020-03-19
Squash whitespace-only string in Makefile.PL to quiet EU::MM

d-lamb:
If both of these variables are empty, then the string here would
be just " ", and that would generate a few 'uninitialized' warnings
during 'perl Makefile.PL'. No major problem, but it looks bad.
This sets it to the empty string in that case.

This is a lot of notes for a 12-character addition but since it potentially effects all builds I wanted to have it looked at before pushing to master.


d-lamb:
The reason for the Travis failure is that the /r option to s/// came in Perl 5.14, and we're testing back to 5.10.

PR #278, opened 2020-05-22
Add imageformat to IO::Pnm::Pic to export an access to chkform

wlmb:
Add a new exportable subroutine, imageformat, to PDL::IO::Pic, to give access to chkformat, so that a program can find out easily the format of an image file and then check if it can be openend before attempting, avoiding a possible failure.


d-lamb:
Thanks for the pull request.  Merged and closed.

PR #281, opened 2020-09-03
Fix wfits so that explicitly setting BITPIX works.

### LINKED ISSUE #280: wfits ($pdl, $file, $bitpix) does not change the image data type
According to the documentation, the three-argument form of wfits() should set the BITPIX card in the written file. It does not.

To get the required behaviour an empty hash must be given as a fourth argument.  Here is a sample script.

my $x = rfits('foo.fits');     # Where foo.fits contains a floating point image, i.e. BITPIX == -32.
$x->wfits('bar.fits', 16);    # bar.fits still contains a floating point image with BITPIX == -32.
$x->wfits('baz.fits', 16, {});    # baz.fits now contains a fixed point image with BITPIX == 16.

is needed.

Either the documentation should be changed or, preferably, the three-argument form implemented correctly by checking for an undef fourth argument.

d-lamb:
This should close #280 .

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,7 +21,7 @@ install:
   - cpan App::cpanminus
   # F77 1.23, which comes with SP 5.30.1.1, fails on Win32
   - cpanm ExtUtils::F77
-  - cpanm -q --showdeps --with-develop --with-suggests --with-configure . | findstr /v "^perl\>" | cpanm -n
+  - cpanm -q --showdeps --with-develop --with-suggests --with-configure . | findstr /v "^perl\>" | findstr /v "^Config" | cpanm -n
   - 'echo End install at: & time /t'
 
 build_script:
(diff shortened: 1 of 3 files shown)


d-lamb:
@zmughal any ideas on these AppVeyor build failures?  Builds 1.0.802 and 1.0.803 both have messages at the end:

> ! Finding Config on cpanmetadb failed.
! Finding Config () on mirror http://www.cpan.org failed.
! Couldn't find module or a distribution Config

Restarting the build doesn't seem to help.  I might wait 24 hours and try again.

shawnlaffan:
It's not an Appveyor thing.  Running ```cpanm --look Configure``` from my command line gives the same error.  

Changing line 24 to also filter Config might work, such that    

```  - cpanm -q --showdeps --with-develop --with-suggests --with-configure . | findstr /v "^perl\>" | cpanm -n```

becomes

```  - cpanm -q --showdeps --with-develop --with-suggests --with-configure . | findstr /v "^perl\>" | findstr /v "^Config" | cpanm -n```

d-lamb:
@shawnlaffan thanks for the pointer.  That seemed to do the trick.

PR #284, opened 2020-09-08
Remove extrapolation code from PDL::GSL::INTERP.

### LINKED ISSUE #279: PDL::GSL::INTERP supports extrapolation when GSL>=1.15 does not
Yury Pakhomov reported on the pdl-general mailing list that PDL::GSL::INTERP was returning NaN's when used for extrapolation.  This is due to an change introduced in GSL 1.15 that "The interpolation functions now return NaN " …  "when x is out of range, instead of extrapolating." (see [the GSL NEWS file](http://git.savannah.gnu.org/cgit/gsl.git/tree/NEWS).)  [This GSL mailing list thread](https://lists.gnu.org/archive/html/help-gsl/2011-12/msg00024.html) and the links there show the change and the train of thought.

Clearly PDL::GSL::INTERP needs to be updated to reflect this reality: extrapolation should not be permitted if GSL version is >=1.15.  Or we could just remove the extrapolation option.

This code shows the error:
```perl
$x=sequence(9)*0.5;
$y=pdl(qw( 0.555  0.566  0.567  0.567  0.569  0.572  0.577  0.584 0.591));

$spl=PDL::GSL::INTERP->init('akima',$x,$y);

$xx=sequence(11)*0.5;
$yy=$spl->eval($xx,{Extrapolate=>1});

print $yy;
[0.555 0.566 0.567 0.567 0.569 0.572 0.577 0.584 0.591 NaN NaN]
```

d-lamb:
Extrapolation seems to have been accidentally allowed by the
GSL interpolation routines. Since GSL v1.15 (released in 2011),
extrapolation returned NaN instead of extrapolating.  These changes
cause the interpolation routines to return an 'input domain error'
if the user attempts to extrapolate.

This should close #279.

### DIFF
--- a/Lib/GSL/INTERP/gsl_interp.pd
+++ b/Lib/GSL/INTERP/gsl_interp.pd
@@ -92,77 +92,64 @@ Example:
 
 =for ref
 
-The function eval returns the interpolating function at a given point. By default
-it will barf if you try to extrapolate, to comply silently if the point to be
-evaluated is out of range pass the option {Extrapolate => 1}
+The function eval returns the interpolating function at a given point.
+It will barf with an "input domain error" if you try to extrapolate.
+
 
 =for usage
 
 Usage:
 
-    $result = $spl->eval($points,$opt);
+    $result = $spl->eval($points);
 
 =for example
 
 Example:
 
     my $res = $spl->eval($x)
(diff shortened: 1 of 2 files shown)


mohawk2:
Great work, thanks!

Issue #288, opened 2020-11-20
bad values from Rice compressed FITS not handled correctly.


pdl-user:
When rfits reads a Rice compressed FITS file, pixels with the value BLANK are not marked as bad.
The code block below generates a 512x480 image with 256 32x30 blocks of values from -32768
to 32767 in steps of 257 and BLANK is set to -32768 . (Code for  gen_stepped_fits.pdl at bottom.) Then the CFITSIO utility imcopy is used to Rice compress and again to decompress.
```bash
$ ./gen_stepped_fits.pdl
$ imcopy stepped.fits stepped_rice.fits[compress]
$ imcopy stepped_rice.fits stepped_unrice0.fits
$ imcopy stepped_rice.fits[1] stepped_unrice.fits
$ pdl
perlDL shell v1.357
 PDL comes with ABSOLUTELY NO WARRANTY. For details, see the file
 'COPYING' in the PDL distribution. This is free software and you
 are welcome to redistribute it under certain conditions, see
 the same file for details.
ReadLines, NiceSlice, MultiLines  enabled
Reading .perldlrc...
Type 'demo' for online demos
Loaded PDL v2.024 (supports bad values)
pdl> $PDL::verbose = 1
pdl> $img = rfits 'stepped.fits'
Reading IMAGE data...
BITPIX =  16  size = 245760 pixels
Reading  491520  bytes
FITS file may contain bad values.
BSCALE =  &&  BZERO =

pdl> p $img->hdr->{BLANK}
-32768
pdl> p $img->min, $img->sum
-32511 31334400
```
So far, so good.

Next stepped_rice.fits is read with ds9 and the lower left corner block is properly displayed as BLANK (bad).
![image](https://user-images.githubusercontent.com/74427119/99737148-8d4ee000-2a7c-11eb-92c3-391bd2b50375.png)
But when stepped_rice.fits is read with PDL:
```
pdl> $img_rice = rfits 'stepped_rice.fits[1]'
Reading 3840 bytes of table data and 177600 bytes of heap data....

pdl> p $img_rice->hdr->{BLANK}
-32768
pdl> p $img_rice->min, $img_rice->sum
-32768 -122880
```
This time no print of:
`FITS file may contain bad values.`
and the minimum and sum are much lower.

When the Rice compressed file is decompressed, everything is OK again.

```
pdl> $img_unrice = rfits 'stepped_unrice.fits'
Reading IMAGE data...
BITPIX =  16  size = 245760 pixels
[...]

PR #289, opened 2020-12-11
rfits properly treats BLANK in rice-compressed images with no BSCALE

d-lamb:
As identified in issue #288, if a Rice-compressed image had a BLANK
keyword in the FITS header, that would not properly get set to BAD
if the FITS header did not also contain the BSCALE keyword.  This
was because treat_bscale(), which also deals with the BLANKs, was
not called if there was no BSCALE present.

Ideally this change would have a corresponding test, but without
compression in wfits, I don't see how to do it simply.

### DIFF
--- a/IO/FITS/FITS.pm
+++ b/IO/FITS/FITS.pm
@@ -1609,6 +1609,8 @@ FITS standard, by adding an option hash to the arguments:
 
 =item compress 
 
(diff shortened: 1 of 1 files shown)


mohawk2:
@d-lamb If you've tried this and you're happy it works, feel free to merge.

PR #295, opened 2021-03-15
Doc complex

fantasma13:
I've added a few snippets of documentation in several places. The two commits can be treated as one. 

There are still fundamental design decisions to be made, I think.
- creal, cimag, cabs, carg should cast onto float/double.
- comparison of complex numbers should fail.
In my opinion.

### DIFF
--- a/Basic/Complex/complex.pd
+++ b/Basic/Complex/complex.pd
@@ -48,6 +48,19 @@ just treat the piddle as a real-valued piddle with a lowest dimension of
 size 2, so C<max> will return the maximum of all real and imaginary parts,
 not the "highest" (for some definition)
 
+=head2 Native complex support
+
+2.027 added changes in complex number handling, with support for C99
+complex floating-point types, and most functions and modules in the core
+distribution support these as well.
+
+PDL can now handle complex numbers natively as scalars. This has
+the advantage that real and complex valued piddles have the same
+dimensions. Consider this when writing code in the future.
+
+See L<PDL::Ops/ci>, L<PDL::Ops/creal>, L<PDL::Ops/cimag>,
+L<PDL::Ops/cabs>, L<PDL::Ops/carg>, L<PDL::Ops/conj> for more.
+
 =head1 TIPS, TRICKS & CAVEATS
 
 =over 4
--- a/Basic/Core/Core.pm
+++ b/Basic/Core/Core.pm
@@ -2823,6 +2823,14 @@ Convert to float datatype
 
 Convert to double datatype
 
+=head2 cfloat
+
+Convert to complex float datatype
+
+=head2 double
+
+Convert to complex double datatype
+
 =head2 type
 
 =for ref
--- a/Basic/Pod/PP.pod
+++ b/Basic/Pod/PP.pod
@@ -2027,8 +2027,8 @@ Only used if C<HandleBad =E<gt> 1>.
 =item GenericTypes
 
 An array reference. The array may contain any subset of the one-character
-strings `B', `S', `U', `L', `Q', `F' and `D', which specify which types
-your operation will accept. The meaning of each type is:
+strings given below, which specify which types your operation will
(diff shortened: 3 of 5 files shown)


Basic/Pod/PP.pod
```diff
@@ -2041,8 +2041,8 @@ Only used if C<HandleBad =E<gt> 1>.
 =item GenericTypes
 
 An array reference. The array may contain any subset of the one-character
-strings `B', `S', `U', `L', `Q', `F' and `D', which specify which types
-your operation will accept. The meaning of each type is:
+strings `B', `S', `U', `L', `Q', `F', `D', 'C' and 'G' which specify which
```
mohawk2, line 2044: Better to delete the list of letters entirely, given they are then listed straight after. This will also reduce the doc effort for adding further types such as [complex] long double.

Basic/Pod/Dataflow.pod
```diff
@@ -173,6 +173,11 @@ All slices or transformations that simply select a subset of the original
 pdl are two-way. Matrix inverse should be. No arithmetic
 operators are.
 
+A note on (native) complex numbers: At the moment, creal, etc. are, in effect, slicing
+operations and not projections with type casts. This means, dataflow is
+active. This is still under discussion and the behaviour may change. Remember
```
mohawk2, line 178: This is currently the opposite of true. `creal` et al just copy out the real or imaginary component.

Basic/Core/Core.pm
```diff
@@ -2821,7 +2821,13 @@ Convert to float datatype
 
 =head2 double
 
-Convert to double datatype
```
mohawk2, line 2824: Please don't delete documentation!

Basic/Complex/complex.pd
```diff
@@ -25,6 +25,34 @@ pp_addpm {At => Top}, <<'EOD';
...
+
+=item conj - complex conjugate
+
+Most exisitng functions work well regardless of input. Some require special handling, e.g. everything
+that is based on cmparison, e.g. max, min, etc. At the moment, the real part is compared. This is still
+under discussion and may change in the future.
+
+=cut
+
+
```
mohawk2, line 55: No extra blank lines please

Basic/Complex/complex.pd
```diff
@@ -25,6 +25,34 @@ pp_addpm {At => Top}, <<'EOD';
...
+=for
+
+=item ci() - returns imaginary unit 
+
+=item creal, cimag, cabs, carg - return real, imaginary, absolute (length) and argument (angle) of 
+complex numbers.
+
+=item conj - complex conjugate
+
+Most exisitng functions work well regardless of input. Some require special handling, e.g. everything
```
mohawk2, line 49: Spelling

mohawk2: Also, are you sure that comparison is done only on the real part? If so, this is a bug and needs fixing. Can you add comments in the discussion of this PR showing it's true?

Basic/Complex/complex.pd
```diff
@@ -25,6 +25,34 @@ pp_addpm {At => Top}, <<'EOD';
...
+has been added, most functions and modules in the core distribution are supporting
+these as well. 
+
+PDL can now handle complex numbers natively as scalars. This has the advantage that
+real and complex valued piddles have the same dimensions. Consider this when writing
+code in the future.
+
+Important functions are:
+
+=for
```
mohawk2, line 40: Are you sure this is the right markup? Is it used somewhere else, or should this actually be `=over`?

mohawk2 (changes requested):
Thank you for having a go at this! There are a few tweaks I've suggested. Also, if you want to squash commits together you can actually do that locally (using `git rebase -i master`) and force-push (with `git push -f`). I would encourage you to do so, in order to develop your git ability.

mohawk2:
I have fixed the above-mentioned issues.

PR #244, opened 2018-06-17
Eliminate $a and $b from docs and code

mohawk2:
This will fix #150.

PR #296, opened 2021-03-21
Change couple of variables in doc

wlmb:
I believe there was a typo, so I renamed two variables $x and $y to $a and $b to make them compatible with the accompanying diagrams in the description of Dataflow


mohawk2:
In fact, any changes should be in the other direction! The reason is to avoid encouraging people to use `$a` and `$b` outside of sorting, because it accidentally avoids `strict`. I'll be pleased to accept renaming other vars to e.g. `$z` so it looks consistent, but we can't have this change as-is. See #244 for the agonising process we went through :-)

wlmb:
Right! So I changed again the variable names and the diagrams using the range u-z, hopefully in a consistent way.

mohawk2:
Thanks! I've changed the original `$x` and `$y` back as well, and also squashed the commits together. A technique you may find helpful (I do this constantly myself) is to review changes locally using `git log -p -w`.

PR #298, opened 2021-04-03
Add tests for mixed PDL::Complex binop cdouble operations

wlmb:
Added a few (TODO) tests for the type and value of PDL::Complex + and  * cdouble

### DIFF
--- a/Basic/Complex/t/complex.t
+++ b/Basic/Complex/t/complex.t
@@ -373,4 +373,19 @@ TODO: {
     ok($zero_imag!=5,'neq real');
 }
 
+# Test binary operations between PDL::Complex and cdouble types.
+TODO: {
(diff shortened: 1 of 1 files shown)


mohawk2:
I've just rebased and force-pushed this, catching up with over 740(!) commits on `master`.

mohawk2:
@wlmb Again, thank you for making these! Since then, my thinking is that PDL::Complex is best not disturbed further, given that users can explicitly use `PDL::Complex->from_native` and `$pdlc->as_native` to go backwards and forwards explicitly. Given the horrors of the extra dimension, my feeling is that is much, much better than hiding the difference.

What do you think? If you agree, we can close this PR.

wlmb:
> 
> What do you think? If you agree, we can close this PR.

I guess we can.

mohawk2:
Thank you!

PR #299, opened 2021-04-17
Remove PDL_clean_namespace and PDL_OLD_API defines

zmughal:
The use of the old PDL API was added in 2002 specifically to backport a
fix for Gimp-Perl <https://github.com/PDLPorters/pdl/commit/cd83ceb6620e36c48b337f2675e7f5048c62e6eb>
so that it could continue to use `PDL->new()` instead of
`PDL->pdlnew()` in the C API.

This makes using that old API an error.

As far as I can tell, Gimp-Perl is the only user of this API. There is
an associated merge request at <https://gitlab.gnome.org/GNOME/gimp-perl/-/merge_requests/1>
which updates Gimp-Perl to the new API.


mohawk2:
Thanks!

PR #307, opened 2021-04-28
GitHub Actions: fix repo constraint for IRC notifications

zmughal:
- Use expression syntax in more places
- Add repo constraint to steps instead of job


zmughal:
Compare with <https://github.com/zmughal/p5-PDL/actions/runs/791208146>.

mohawk2:
Fingers crossed. And thanks as always for your dedication!

PR #310, opened 2021-04-30
CI: only run Cygwin build on tag creation (at release time)

zmughal:
Cygwin CI builds run slowly at the moment and it is not necessary to run
tests on Cygwin for every commit. This will still help catch problems
while not interfering with other builds.


mohawk2:
Thanks, this is great!

PR #333, opened 2021-08-08
Use shorter commit message in IRC

zmughal:
So that the IRC log is less cluttered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,11 @@ jobs:
       IRC_PORT: '6667'
       IRC_NICKNAME: 'pdl-commits'
     steps:
+      - uses: actions/checkout@v2
+      - name: irc push (build messages)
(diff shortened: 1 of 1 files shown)


mohawk2 (approved):
I'm not quite sure I understand why add `actions/checkout` - if it's to populate `github.event.commits.*.id`, then that's great.

In any case, I have faith, so if you're happy, merge away!

zmughal:
On 2021-08-07 at 20:56:41 -0700, mohawk2 wrote:
> @mohawk2 approved this pull request.
> 
> I'm not quite sure I understand why add `actions/checkout` - if it's to populate `github.event.commits.*.id`, then that's great.

The `actions/checkout` is needed for setting `env.commitmsg` using `git
show` using `github.event.commits.*.id`. Otherwise if we use the
messages directly in `github.event.commits.*.message`, these contain the
entire commit message instead of the first line.

> In any case, I have faith, so if you're happy, merge away!

Will do!

PR #341, opened 2021-10-04
minor speedups for _establish_type

shawnlaffan:
When a scalar variable is not a ref then there is no need to call _establish_type to only get the default value back, particularly where this will always be <= the type found so far.

Passes all tests but perhaps there are some edge cases lurking.

### DIFF
--- a/Basic/Core/Core.pm
+++ b/Basic/Core/Core.pm
@@ -1157,7 +1157,7 @@ sub PDL::new {
        }
        $value = ($PDL::undefval//0)+0
    }
-   $type = _establish_type($value, $PDL_D) if !defined $type;
+   $type //= ref($value) ? _establish_type($value, $PDL_D) : $PDL_D;
 
    return pdl_avref($value,$this,$type) if ref($value) eq "ARRAY";
    my $new = $this->initialize();
--- a/Basic/Core/Core.pm
+++ b/Basic/Core/Core.pm
@@ -1142,7 +1142,9 @@ sub _establish_type {
   return $PDL_CD if UNIVERSAL::isa($item, 'Math::Complex');
   return max($item->type->enum, $sofar) if UNIVERSAL::isa($item, 'PDL');
   return $PDL_D if ref($item) ne 'ARRAY';
-  max($sofar, map _establish_type($_, $sofar), @$item);
+  #  only need to recurse for items that are refs
+  #  as $sofar will be $PDL_D at a minimum
+  max ($sofar, map {_establish_type($_, $sofar)} grep {ref} @$item);
 }
 
 sub PDL::new {


Basic/Core/Core.pm
```diff
@@ -1142,7 +1142,9 @@ sub _establish_type {
   return $PDL_CD if UNIVERSAL::isa($item, 'Math::Complex');
   return max($item->type->enum, $sofar) if UNIVERSAL::isa($item, 'PDL');
   return $PDL_D if ref($item) ne 'ARRAY';
-  max($sofar, map _establish_type($_, $sofar), @$item);
+  #  only need to recurse for items that are refs
+  #  as $sofar will be $PDL_D at a minimum
+  max ($sofar, map {_establish_type($_, $sofar)} grep {ref} @$item);
```
mohawk2, line 1147: Please don't change the `map thing($_), @list` syntax back to `map {thing($_)} @list`, it adds an unnecessary scope-creating Perl op which slows things down. The same syntax is available for `grep` too.

mohawk2 (changes requested):
I like the principle, but please add a before-and-after performance comparison if you are claiming to improve performance :-)

Basic/Core/Core.pm
shawnlaffan: Done in b57df43c2c2685f9aeb07ce98e9f0219b7417e7f

mohawk2: Thanks!

mohawk2:
Thanks for addressing the syntax point! How about the actual performance benchmarking? :-)

shawnlaffan:
Some simple benchmark results and code are below.   Was about to post with latest commit but got distracted by other matters.  

Array processing is faster.  Scalar assignment is no different, which is not surprising given it's only one sub call that's skipped.  

I'm not sure how well the complex subs represent common usage but it does show they are no different.  

I'm also not sure what's going on with the result of the native complex nelem call, but it's not affected by this PR.  



```
C:\git\pdl>perl bench_pdl_new.pl
Using Perl version 5.032001
Using PDL version 2.057
Number of elements and type:
array:       500000 double
scalar:      1 double
complex:     300 cdouble
complex2:    300 cdouble
complex_pdl: 3628800 cdouble
Benchmark: running array, complex, complex2, cplx_pdl, scalar for at least 3 CPU seconds...
     array: 3.25779 wallclock secs ( 3.20 usr +  0.01 sys =  3.22 CPU) @  6.21/s (n=20)
   complex: 3.20162 wallclock secs ( 3.20 usr +  0.00 sys =  3.20 CPU) @ 7735.56/s (n=24777)
  complex2: 3.16837 wallclock secs ( 3.11 usr +  0.02 sys =  3.12 CPU) @ 78128.96/s (n=244153)
  cplx_pdl: 3.28951 wallclock secs ( 2.27 usr +  1.00 sys =  3.27 CPU) @ 27.86/s (n=91)
    scalar: 3.01108 wallclock secs ( 3.00 usr +  0.00 sys =  3.00 CPU) @ 199913.70/s (n=599941)

C:\git\pdl>perl -Mblib bench_pdl_new.pl
Using Perl version 5.032001
Using PDL version 2.057_03
Number of elements and type:
array:       500000 double
scalar:      1 double
complex:     300 cdouble
complex2:    300 cdouble
complex_pdl: 3628800 cdouble
Benchmark: running array, complex, complex2, cplx_pdl, scalar for at least 3 CPU seconds...
     array: 3.07567 wallclock secs ( 2.89 usr +  0.12 sys =  3.02 CPU) @ 33.82/s (n=102)
   complex: 3.16486 wallclock secs ( 3.14 usr +  0.00 sys =  3.14 CPU) @ 18854.46/s (n=59203)
  complex2: 3.22989 wallclock secs ( 3.23 usr +  0.00 sys =  3.23 CPU) @ 73232.84/s (n=236835)
  cplx_pdl: 3.20334 wallclock secs ( 2.23 usr +  0.91 sys =  3.14 CPU) @ 25.14/s (n=79)
    scalar: 3.4221 wallclock secs ( 3.39 usr +  0.00 sys =  3.39 CPU) @ 196998.53/s (n=668022)
```


```perl
use strict;
use warnings;
use 5.010;

use Math::Complex qw//;
use PDL;
#use PDL::NiceSlice;

use Benchmark qw /timethese :hireswallclock/;

say "Using Perl version $]";
say "Using PDL version $PDL::VERSION";

my $array_data = [([1..10000]) x 50];
my $mc_data = [([1..100]) x 3];
$mc_data->[1][9]  = Math::Complex::cplxe(1, 2*Math::Complex::pi/3);
$mc_data->[2][15] = Math::Complex::cplxe(1, 2*Math::Complex::pi/4);
my $mc2_data;
push @$mc2_data, Math::Complex::cplx(1, 2 * Math::Complex::pi/3);
push @$mc2_data, Math::Complex::cplx(1, 2 * Math::Complex::pi/4);
my $cplx_data = i (1..10);

my $from_scalar      = sub {PDL->new (5);};
my $from_array       = sub {PDL->new ($array_data);};
my $from_complex     = sub {PDL->new ($mc_data);};
my $from_complex2    = sub {PDL->new ($mc2_data);};
my $from_complex_pdl = sub {PDL->new ($cplx_data);};


say 'Number of elements and type:';
my $x;
$x = $from_array->();   say 'array:       ', $x->nelem, ' ', $x->type;
$x = $from_scalar->();  say 'scalar:      ', $x->nelem, ' ', $x->type;
$x = $from_complex->(); say 'complex:     ', $x->nelem, ' ', $x->type;
$x = $from_complex->(); say 'complex2:    ', $x->nelem, ' ', $x->type;
$x = $from_complex_pdl->();  say 'complex_pdl: ', $x->nelem, ' ', $x->type;


timethese(-3, {
    'scalar'   => $from_scalar,
    'array'    => $from_array,
    'complex'  => $from_complex,
    'complex2' => $from_complex2,
    'cplx_pdl' => $from_complex_pdl,
});
```

shawnlaffan:
Scalar is likely within timing noise.  A scalar assignment should take less time than a full sub call with all its associated checks (as fast as they might be).

The complex variants can be made faster in a couple of ways, depending on the expected common pattern of values being passed.

If it is rare that users pass mixed arrays containing both normal and complex vals (native or Math::Complex) then an additional check can be added before the map, calling _establish_type on the first value in the array.  If the value is complex then the sub can return early since it is of the max type.  This will slightly slow the PR's processing of "vanilla" arrays due to the extra sub call but it should still be faster than the current implementation on CPAN (one _establish_type call vs many).

If it is common that users pass mixed arrays then the map/grep can be replaced by a for loop that skips iters where the value is not a ref, calls _establish_type if it is a ref, and exits early if the max type is returned.  This will be slightly slower for the "vanilla" array case than the map/grep approach.  (This is also the first approach I used, but is a larger code change than adding the grep).  


I would guess that the common usage is for arrays with consistent ref types (i.e. all vanilla scalars or all complex), but you'll have a better sense of that than I do.  I'll update the PR once I know.

shawnlaffan:
Thanks for the reviews, and for merging.

PR #347, opened 2021-11-11
Test that $PDL::Core::VERSION matches $PDL::VERSION when first set

shawnlaffan:
This will help avoid red herrings in prereq reporting tests
such as https://www.cpantesters.org/cpan/report/b610a1c0-6bf7-1014-a048-271994176946
where PDL::Core is reported as 2.028 instead of 2.057.

Possibly should be an author test under the xt dir.  

Could also be extended to other files with hard coded versions if a common version is deemed useful, but PDL::Core seems to be the only file that updates its version after it is initially set.

### DIFF
--- a/Basic/Core/Core.pm
+++ b/Basic/Core/Core.pm
@@ -8,7 +8,7 @@ use PDL::Exporter;
 require PDL; # for $VERSION
 use DynaLoader;
 our @ISA    = qw( PDL::Exporter DynaLoader );
-our $VERSION = '2.028'; # PAUSE insists - below is the real one
+our $VERSION = '2.060'; # PAUSE insists - below is the real one
 $VERSION = $PDL::VERSION;
 bootstrap PDL::Core $VERSION;
 use PDL::Types ':All';
--- a/MANIFEST
+++ b/MANIFEST
@@ -716,6 +716,7 @@ Perldl2/TODO
 README
 t/00-report-prereqs.t
 t/01-pptest.t
+t/02-pdl-core-version.t
 t/autoload.t
 t/bad.t
 t/basic.t
--- a/t/02-pdl-core-version.t
+++ b/t/02-pdl-core-version.t
@@ -1,0 +1,22 @@
+#  logic adapted from 00-report-prereqs.t
+
+use strict;
+use warnings;
+
+use Test::More;
+
+use ExtUtils::MakeMaker;
+use File::Spec;
+use PDL;
+
+my $file = 'PDL::Core';
+$file =~ s{::}{/}g;
+$file .= ".pm";
+my ($prefix) = grep { -e File::Spec->catfile($_, $file) } @INC;
+
+my $filename         = File::Spec->catfile($prefix, $file);
+my $pdl_core_version = MM->parse_version( $filename );
+
+is $pdl_core_version, $PDL::VERSION, 'versions match: PDL and PDL::Core';
+
+done_testing;


mohawk2:
Thank you for the contribution! (though see below)

PDL::Core is accurately reported because that is that module's version. The only version that is important is correctly reported a couple of lines above, which is PDL 2.057. That tester setup is broken in some way, because it shouldn't be getting a conflict between the PDL used to build the module, and that used to test it.

I changed the PDL module versions a little while ago (in fact, around 2.028), to make the release process a bit less of a faff. No problems have yet arisen from this. Therefore, I will not be accepting this PR. Please do ask questions another time, since this must have taken you some effort which I would have wanted to save you! Our IRC channel would be one way, or opening an issue, or mailing the mailing list.

shawnlaffan:
No worries.  It did not take long to implement.  

The issue is not one of compilation or use.  However, it does cause a bit of cognitive dissonance when the prereq scanner test in, for example, PDL::Stats reports it is using an old version when in fact it is the current version.

The root cause is that the static scanning used by ```MM->parse_version``` gets the placeholder value and this value is then reported by the prereq test file.  As PDL versions increase this version difference will become further out of step, becoming more of a red herring.

mohawk2:
> No worries. It did not take long to implement.
> 
> The issue is not one of compilation or use. However, it does cause a bit of cognitive dissonance when the prereq scanner test in, for example, PDL::Stats reports it is using an old version when in fact it is the current version.
> 
> The root cause is that the static scanning used by `MM->parse_version` gets the placeholder value and this value is then reported by the prereq test file. As PDL versions increase this version difference will become further out of step, becoming more of a red herring.

The remedy for that would be for CPAN distros to not depend on the wrong modules, and/or for those reading the output of those tests to be aware of the situation. :-)

Are you happy for me to close this PR?

shawnlaffan:
There's no issue with the CPAN deps - it's an issue with the prereq reporting test reporting a version number that differs from what it is in practice.  The hard coded value is overridden by the next line of code to match the current PDL release, but the MM method reports the hard coded value.

Feel free to close, though.

mohawk2:
> There's no issue with the CPAN deps - it's an issue with the prereq reporting test reporting a version number that differs from what it is in practice. The hard coded value is overridden by the next line of code to match the current PDL release, but the MM method reports the hard coded value.

As the person who added the prereq-reporting test, and the current releaser of PDL, I am aware of the above, and comfortable with the tradeoff between work, and the risk of slight confusion.

> Feel free to close, though.

Thank you! On a point not directly connected to this, but a remaining issue; how do you feel about having a go at updating the PDL use of Proj (the two modules within `Libtmp`) to a current version, so we can stop using the version-fixed Alien::Proj4, and switch to using your excellent Alien::proj?

shawnlaffan:
> On a point not directly connected to this, but a remaining issue; how do you feel about having a go at updating the PDL use of Proj (the two modules within Libtmp) to a current version, so we can stop using the version-fixed Alien::Proj4, and switch to using your excellent Alien::proj?

I can have a look but XS bindings are outside my skill set.  We might also be able to do something using [Geo::LibProj::FFI](https://metacpan.org/pod/Geo::LibProj::FFI) though.

mohawk2:
XS bindings aren't a problem. All that's needed is a steer on the changes needed in the C code in the 4 `pp_def` functions in https://github.com/PDLPorters/pdl/blob/master/Libtmp/GIS/Proj/Proj.pd, to take it from the Proj 4 API, to the current one (version 8).

Whether it's done by compiled C code or via FFI, we'll still need the guidance on what C function-calls need updating. My prediction is it won't take much more than a read of the change-log for versions 5 through 8. But it's easier if done by an expert :-)

mohawk2:
In fact, I wouldn't be surprised if the notes at https://github.com/OSGeo/PROJ/releases/tag/5.0.0 aren't most of the answer (apparently the new functionality and idiom is stuff in `proj.h`).

PR #351, opened 2021-11-19
Primitive::srand() added - fix #350

### LINKED ISSUE #350: random() should call srand() implicitly
Tested with PDL 2.051 on Windows and 2.019 on Linux
`>perl -MPDL -E"say random(5)"
[3.907985e-14 0.00098539467 0.041631002 0.17664264 0.36460225]`
will always produce the same results when called repeatedly. It will also give the same results on Windows and Linux. PDL docs for random() says "You can use the perl function srand to seed the random generator. For further details consult Perl's srand documentation." And Perl docs for srand say "If srand() is not called explicitly, it is called implicitly without a parameter at the first use of the rand operator." Therefore I'd expect PDL to also call srand() implicitly at the first use of random(), grandom() etc.

mohawk2:
This fixes #350.


zmughal:
Looks good to me!

PR #352, opened 2021-11-19
Add pdl_seed() function that uses PID and epoch time

zmughal:
References:

  - Helmut G. Katzgraber. "Random Numbers in Scientific Computing:
    An Introduction". <https://arxiv.org/abs/1005.4117v1>.
  - `Perl_seed()` in the Perl source code.

### DIFF
--- a/Basic/Core/Core.xs
+++ b/Basic/Core/Core.xs
@@ -212,6 +212,13 @@ _inf(...)
 
 MODULE = PDL::Core     PACKAGE = PDL::Core
 
+IV
+seed()
+  CODE:
+    RETVAL = pdl_pdl_seed();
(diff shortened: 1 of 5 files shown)


mohawk2:
Looks really good (the CI shows tweaks are needed)! I believe the best location for the C code would be in `pdlcore.c`, and the declaration should be on the end of `PDL_CORE_LIST` (the end so it's binary-compatible) in `pdlcore.h`. Then it could be called with `PDL->seed()` in the `PDL_MAYBE_SRAND`, and then a little glue XS function in `Core.xs` could be called from `PDL::srand`.

mohawk2:
Thanks! Merged and released as 2.063.

PR #360, opened 2022-01-04
Use File::Which to find executables

zmughal:
`File::Which::which()` replaces the ad hoc `inpath()` functions.


mohawk2 (approved):
Thanks!

PR #368, opened 2022-01-11
Fix POD errors.

sebastic:
The lintian QA tool reported a POD error for the Debian package build:
```
 Around line 1413:
  alternative text '/axis option' contains non-escaped | or /
```


mohawk2:
Thank you!

Issue #370, opened 2022-01-13
sequence() fails as instance method


moocow-the-bovine:
Since commit f49ac06762650892960c2e1fa9ab7345ba31f039 (v2.063_03), sequence() called as an instance method returns a copy of the calling object whenever its type is not `double`.

Examples:
```perl
$vals = [3,4,5];
$good = pdl(double, $vals)->sequence;  # ok: [0 1 2]
$bad = pdl(float, $vals)->sequence; # bad: [3 4 5]
$bad = pdl(indx, $vals)->sequence; # bad: [3 4 5]
# etc.
```

This leads to downstream failures in PDL::CCS (-> https://github.com/moocow-the-bovine/PDL-CCS/issues/1), and appears to have resulted from changes made to `PDL::sequence()` and `PDL::axisvals2()` in the aforementioned commit.  I'm guessing that the changed behavior is related to the new `$type_given` lexical in `PDL::sequence()`  (which is false when called as an instance method), but I don't have a fix.

`PDL::axisvals2()` appears to be generating the correct values (in both `$bar` and `$dummy`), albeit always as type `double` (which seems rather inefficient).  Perhaps the `inplace()` on the `sequence()` lexical `$bar` isn't propagating correctly through `axisvals2()` when there's a type-mismatch?

mohawk2:
Thank you for figuring out where PDL::CCS's expectation was being broken! I'll add this to the unit test and then see about fixing it.

mohawk2:
In fact, thank you for the unit tests as well! That saves time :-D

mohawk2:
FYI I've moved the tests from t/core.t to t/basic.t, because despite how fundamental `sequence` is in PDL programming, it's a implemented in PDL::Basic rather than PDL::Core. Yes, those packages want a good tidy-up in due course.

mohawk2:
By the way, `inplace` for different types is semantically a bit confusing. I saw in CCS that you've got a `convert` that works inplace, which is something I'm looking at enabling in main PDL. The main problem it creates is when there are badvalues - if you change the data in-place (which is itself a problem - what happens when you convert `byte` to `double`?), what should the new ndarray's (or supposedly, due to `inplace`, the same one) badvalue be? If you have insights into how to make this work well, I'm very interested!

Another issue is that in the hopefully-near-future loop-fusion change, the whole concept of `inplace` will become a bit moot, and might actually get ignored. Ideas on this are extremely welcome!

mohawk2:
[...]

PR #371, opened 2022-01-13
tests for sequence() as instance method

moocow-the-bovine:
add tests for sequence-as-instance method bug described in https://github.com/PDLPorters/pdl/issues/370


mohawk2:
As I only had to add one line of code to make this pass, I left it as yours. As a side-benefit, this means the CI continues to show `master` as passing, which is the standard we aim for.

PR #372, opened 2022-01-18
Improve version parsing in some GSL modules

d-lamb:
When `gsl-config --version` returns a dotted triple like '2.7.1',
simple parsing like `$version <2.0` does not work.


mohawk2:
Thanks! I was just noticing these errors and was going to have a go.

PR #378, opened 2022-01-28
Improve check for packing quad values

zmughal:
- Check if quad values can be packed by using `eval`. This is because
  the condition for whether `pack` can use quad values is a bit more
  complicated. See:

    * <https://www.nntp.perl.org/group/perl.perl5.porters/2013/10/msg208380.html>,
    * <https://github.com/Perl/perl5/commit/1640b983ff7c1296873ae4dc0e2163160c853cb3>,
    * <https://github.com/Perl/perl5/commit/c174bf3bd0aaba83d7aba81b0fbcb0055109a168>

- Add unsigned quad value (Q) to the condition.

### DIFF
--- a/Basic/Core/Core.pm
+++ b/Basic/Core/Core.pm
@@ -16,12 +16,11 @@ use Config;
 use List::Util qw(max);
 use Scalar::Util 'blessed';
 
+# If quad (q/Q) is available for pack().
+our $CAN_PACK_QUAD = !! eval { my $packed = pack "Q", 0; 1 };
+
 # If "D" is available for pack().
-# See commit <https://github.com/Perl/perl5/commit/77a2054ea312b030904555167b764e4510dbcac6>
-# and PR <https://github.com/Perl/perl5/pull/18517>.
-our $CAN_PACK_D =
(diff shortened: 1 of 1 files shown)


Basic/Core/Core.pm
```diff
@@ -1204,7 +1206,7 @@ sub PDL::new {
          # new was passed a string argument that doesn't look like a number
          # so we can process as a Matlab-style data entry format.
 		return PDL::Core::new_pdl_from_string($new,$value,$this,$type);
-      } elsif ($Config{ivsize} < 8 && $pack[$new->get_datatype] eq 'q*') {
+      } elsif (! $CAN_PACK_QUAD && $pack[$new->get_datatype] =~ /^[qQ]\*$/ ) {
```
mohawk2, line 1209: This could be `/q/i`?

mohawk2: I think you're right, let's also simplify `PACK_D` to use `eval` too.

mohawk2 (approved):
Subject to simplifying as you already said, looks great! Thank you for jumping on this.

Basic/Core/Core.pm
zmughal: Done and done!

PR #387, opened 2022-02-21
GHA: Add workflow for downstream testing

zmughal:


zmughal:
@djerius, @ebaudrez, @fantasma13, @jjatria, @jlapeyre, @kmx, @lab-measurement (@amba, @akhuettel), @moocow-the-bovine, @mvgrimes, @run4flat, @sergeykolychev, @shawnlaffan, @Starlink (@grahambell), @stphnlyd, @wlmb

This PR tests a distribution of yours against the latest PDL (very much like Perl "blead" is tested against CPAN). You do not have to do anything now (though in the future there may be some bug reports/PRs filed to make sure that your code continues to work).

For a list of which modules are tested, see this YAML file: <https://github.com/PDLPorters/devops/blob/master/data/project.yml>. If you have more dists you would like to submit for downstream testing, let me know!

PR #389, opened 2022-03-08
Clarify reason for not building TriD graphics.

d-lamb:
This separates out the two module requires and version checks, and reports to the user the reason that TriD graphics will not be built.


mohawk2:
Thanks!

PR #410, opened 2022-09-27
Add complex support to interpolate(), fixes #408

### LINKED ISSUE #408: `interpolate` and `interpol` drop complex values
```perl
pdl> $x = sequence(10);
pdl> $xi = sequence(10) + 0.5
pdl> $y = sequence(10) + sequence(10)*i()
pdl> p $y
[0 1+i 2+2i 3+3i 4+4i 5+5i 6+6i 7+7i 8+8i 9+9i]
pdl> ($yi, $err) = interpolate($xi, $x, $y); 
pdl> p $yi
[0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5]
```
The last line should have complex values.

KJ7LNW:
This fixes #408.  Not sure if this is the "right" way to fix this, so please review:

About types signatures and `complex`, [PDL::PP](https://metacpan.org/pod/PDL::PP) says _"...If the operation is already complex, there is no effect; if not, the output will be promoted to the type's ["complexversion" in PDL::Type](https://metacpan.org/pod/PDL::Type#complexversion), which defaults to cfloat. Note this is controlled both by the [PDL::Types](https://metacpan.org/pod/PDL::Types) data, and the code in [PDL::PP](https://metacpan.org/pod/distribution/PDL/Basic/Pod/PP.pod)."_

What does _"will be promoted to the type's ["complexversion"](https://metacpan.org/pod/PDL::Type#complexversion)"_ mean?

So if `$y` is complex then `$yi` will be too...but if the caller passes `$y` as real then I would think `interpolate()` should return `$yi` as real, too.  Would it?

### DIFF
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2,6 +2,7 @@ use strict;
 use warnings;
 use PDL::Types qw(ppdefs_all types);
 my $F = [map $_->ppsym, grep $_->real && !$_->integer, types()];
+my $AF = [map $_->ppsym, grep !$_->integer, types];
 
 pp_addpm({At=>'Top'},<<'EOD');
 use strict;
@@ -2962,8 +2963,17 @@ for my $func ( [
 pp_def('interpolate',
        HandleBad => 0,
        BadDoc => 'needs major (?) work to handles bad values',
-	Pars => 'xi(); x(n); y(n); [o] yi(); int [o] err()',
-	GenericTypes => $F, # too restrictive ?
+	Pars => 'real xi(); real x(n); y(n); [o] yi(); int [o] err()',
+	GenericTypes => $AF,
+	PMCode => 'sub PDL::interpolate {
+			my ($xi, $x, $y, $yi, $err) = @_;
+			croak "x must be real" if (ref($x) && ! $x->type->real);
+			croak "xi must be real" if (ref($xi) && ! $xi->type->real);
+			$yi //= PDL->null;
+			$err //= PDL->null;
+			PDL::_interpolate_int($xi, $x, $y, $yi, $err);
+			($yi, $err);
+		}',
 	Code => '
 		 $GENERIC() d;
 		 PDL_Indx n  = $SIZE(n);
@@ -3025,6 +3035,9 @@ which is otherwise 0.
 See also L</interpol>, which uses the same routine,
 differing only in the handling of extrapolation - an error message
 is printed rather than returning an error ndarray.
+
+Note that C<interpolate> can use complex values for C<$y> and C<$yi> but
+C<$x> and C<$xi> must be real.
 
 =cut
 
--- a/Changes
+++ b/Changes
@@ -5,6 +5,7 @@
 - change PP-generated XS PROTOTYPES to DISABLE
 - allow [o] on OtherPars (#362)
 - fix conv2d to allow complex arguments (#407) - thanks @wlmb
+- Add complex support to interpolate (#408) - thanks @KJ7LNW
 
 2.080 2022-05-28
 - make IO::STL work on big-endian systems (#394) - thanks @sebastic for report
--- a/t/primitive.t
+++ b/t/primitive.t
@@ -423,7 +423,15 @@ ok(all( $z==pdl(0,0,0,0,1,4,4) ), "one2nd z");
 my $yvalues =  (new PDL( 0..5))   - 20;
 my $xvalues = -(new PDL (0..5))*.5;
 my $x = new PDL(-2);
-is( $x->interpol($xvalues,$yvalues), -16 );
+is( $x->interpol($xvalues,$yvalues), -16, "interpolate: real-valued" );
+}
+
+{
+my $yvalues =  ((new PDL( 0..5))   - 20) * (1+i()) ;
+my $xvalues = -(new PDL (0..5))*.5;
+my $x = new PDL(-2);
+is( $x->interpol($xvalues,$yvalues), -16 - 16*i, "interpolate: complex-valued" );
+ok( !eval { $x->interpol($xvalues*i(),$yvalues) } , "interpolate: x must be real" );
 }
 
 # Some of these tests are based upon those in Chapter 5 of Programming


KJ7LNW:
@zmughal, @mohawk2, others:

```
pdl> $x = sequence(10);
pdl> $xi = pdl[0.7]
pdl> $y = (sequence(10)+5)
pdl> ($yi, $err) = interpolate($xi, $x, $y); 
pdl> print $yi
[5.7]
pdl> help $y
This variable is Double D [10]                 P            0.08KB
pdl> help $yi
This variable is CDouble D [1]                  P            0.02KB
```

So as you can see the input `$y` is a `Double` and the output `$yi` is a `CDouble`, which is of course what the `PDL::PP` docs say will happen when placing `complex` in `Pars` as I've done in this patch.

**Question:** 

- Are there other functions where PDL takes a `Double` for input but always returns a `CDouble`?
    - When (if ever) is this acceptable?
- Is there a way to write the signature so `interpolate()` will return a `Double` if the input is a `Double` and return a `CDouble` iif the input is a `CDouble`?
    - ...or would this require a "complex" version of the `interpolate()` function such as `cinterpolate()`?

mohawk2:
The `real` and `complex` type-modifiers were motivated by, and are used in, PDL::FFTW3 (https://metacpan.org/pod/PDL::FFTW3#rNfftX-(rNfft1,-rNfft2,-rNfft3,-...,-rNfftn)). Their use is always "acceptable" where they are mathematically meaningful and appropriate.

Your current patch doesn't have docs or tests and needs both. I also think it is the wrong way round, as currently a real input will have various of its `Pars` converted into native-complex data. Instead, you should do this, which is one of the options I gave on #408:

```perl
# at the top, from ops.pd
my $AF = [map $_->ppsym, grep !$_->integer, types];
# ...
	GenericTypes => $AF,
	Pars => 'real xi(); real x(n); y(n); [o]yi(); int [o] err()',
```

mohawk2:
This breaks 3 downstream modules, which will need fixing.

KJ7LNW:
@mohawk2, thanks for the hint!

I've pushed a new commit, please review and approve the workflow so we can check coverage.

mohawk2:
Thanks for the update!

While the mighty @zmughal is working on a fix for Types::PDL (which was apparently broken by the new Type::Tiny), this change also needs the addition of some `PMCode` that dies if the `xi` or `x` params are complex; this is needed because the PP code (and core code) won't check for that, but will instead just quietly convert those inputs if complex to real by dropping the imaginary parts. That would then need an additional test, and doc.

Also, the `GenericTypes` are now the right amount of restrictive, so that comment can go :-)

KJ7LNW:
Hi @mohawk2,

I looked at PMCode but...I don't get it: https://metacpan.org/dist/PDL/view/Basic/Pod/PP.pod#PMCode

Would you please write what you think the PMCode should look like, and post it here so I can write tests for it?

These are the things that you would need to present by example so I can understand it:
 - Show how to test the types to see if they are complex.
 - Show how to handle multiple outputs in the PMCode's return value(s)

This is what I've got so far, but as you can see, its rather trivial, and it doesn't even work:

```perl
PMCode => 'sub PDL::interpolate {
    return PDL::_interpolate_int(@_);
  }',
```

... for which `make test` produces:

```
t/func.t ....... 1/? Usage: PDL::_interpolate_int(xi, x, y, yi, err) at ../../blib/lib/PDL/PP.pm line 952.
```

KJ7LNW:
> So, to help me help you, what is it you think the docs should explain more? Perl-function arguments? PDL operation signatures? Optional arguments? Return values?

The previous message on this thread covers what I'm looking for, this is the primary issue:

> 1.  Show how to test the types to see if they are complex.

KJ7LNW:
I found `->type`, will see if checking against `cdouble` works...

mohawk2:
`->type` returns a PDL::Type object, see https://metacpan.org/pod/PDL::Types#PDL::Type-OBJECTS for available methods. It's the `real` method you want (returns a boolean on whether the type is real-only), but that's not documented which needs fixing.

KJ7LNW:
@mohawk2, do you know what this means?

## The error 
```
Error in PDL::Slices::_clump_int: input parameter 'PARENT' is null at ufunc.pd line 565.
	PDL::or(PDL=SCALAR(0x5624259d80b8)) called at primitive.pd line 3077
	PDL::Primitive::interpol(PDL=SCALAR(0x5624259d7db8), PDL=SCALAR(0x5624259dc790), PDL=SCALAR(0x5624259dc8b0)) called at t/primitive.t line 426
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 139.
```

### The PMCode:

```perl
PMCode => 'sub PDL::interpolate {
                my ($xi, $x, $y, $yi, $err) = @_;

                croak "x must be real" if (ref($x) && ! $x->type->real);
                croak "xi must be real" if (ref($xi) && ! $xi->type->real);


                $yi //= PDL->null;
                $err //= PDL->null;

                PDL::_interpolate_int($xi, $x, $y, $yi, $err);

                ($yi, $err);
        }',
```

### ./Basic/Ufunc/ufunc.pd:565
```perl
=for bad

This routine handles bad values.

=cut

*$name = \\&PDL::$name;
sub PDL::$name {
        my(\$x) = \@_; my \$tmp;
        \$x->clump(-1)->${func}( \$tmp=PDL->nullcreate(\$x) ); # <<< line 565
        \$tmp;
}
EOD
```

### t/primitive.t line 426:
```perl
{
my $yvalues =  (new PDL( 0..5))   - 20;
my $xvalues = -(new PDL (0..5))*.5;
my $x = new PDL(-2);
is( $x->interpol($xvalues,$yvalues), -16, "interpolate: real-valued" );
# ^^ previous line is line 426 ^^
}
```

### Basic/Primitive/primitive.pd:3077
```perl
# kept in for backwards compatability
sub interpol ($$$;$) {
    my $xi = shift;     
    my $x  = shift;     # this is line 3077: Why does the shift fail?
    my $y  = shift;
    my $yi = @_ == 1 ? $_[0] : PDL->null;
    interpolate( $xi, $x, $y, $yi, my $err = PDL->null );
    print "some values had to be extrapolated\n"
        if any $err;
    return $yi if @_ == 0;
} # sub: interpol()
*PDL::interpol = \&interpol;
```

### Here is the diff to add PMCode, pretty simple so far:

```diff
diff --git a/Basic/Primitive/primitive.pd b/Basic/Primitive/primitive.pd
index 60503077..9d5f228e 100644
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2964,7 +2964,21 @@ pp_def('interpolate',
        HandleBad => 0,
        BadDoc => 'needs major (?) work to handles bad values',
        Pars => 'real xi(); real x(n); y(n); [o] yi(); int [o] err()',
-       GenericTypes => $AF, # too restrictive ?
+       GenericTypes => $AF,
+       PMCode => 'sub PDL::interpolate {
+                       my ($xi, $x, $y, $yi, $err) = @_;
+
+                       croak "x must be real" if (ref($x) && ! $x->type->real);
+                       croak "xi must be real" if (ref($xi) && ! $xi->type->real);
+
+                       
+                       $yi //= PDL->null;
+                       $err //= PDL->null;
+
+                       PDL::_interpolate_int($xi, $x, $y, $yi, $err);
+
+                       ($yi, $err);
+               }',
        Code => '
                 $GENERIC() d;
                 PDL_Indx n  = $SIZE(n);
diff --git a/t/primitive.t b/t/primitive.t
index 2981bf3a..0aa3a155 100644
--- a/t/primitive.t
+++ b/t/primitive.t
@@ -431,6 +431,7 @@ my $yvalues =  ((new PDL( 0..5))   - 20) * (1+i()) ;
 my $xvalues = -(new PDL (0..5))*.5;
 my $x = new PDL(-2);
 is( $x->interpol($xvalues,$yvalues), -16 - 16*i, "interpolate: complex-valued" );
+ok( !eval { $x->interpol($xvalues*i(),$yvalues) } , "interpolate: x must be real" );
 }
 
 # Some of these tests are based upon those in Chapter 5 of Programming

```

KJ7LNW:
@mohawk2 nevermind, I must not have saved something because its working now...

mohawk2:
The Ufunc is `or` as you can see in the stack trace. `any` is a synonym. It is invalid to call it with a `null`, which is evidently what you are doing, and why you see that error.

KJ7LNW:
@mohawk2, this passed `make test` on my system, give it a spin.

PR #412, opened 2022-10-11
Refactor curses detection to use ncurses*-config

zmughal:
This helps on systems where additional libraries are needed in order to
link against `ncurses`. In particular, this is useful for MSYS2.

### DIFF
--- a/IO/Browser/Makefile.PL
+++ b/IO/Browser/Makefile.PL
@@ -3,6 +3,7 @@ use warnings;
 use ExtUtils::MakeMaker;
 use File::Spec;
 eval { require Devel::CheckLib; Devel::CheckLib->import; };
+eval { require File::Which; };
 
 my @pack = (["browser.pd", qw(Browser PDL::IO::Browser)]);
 
@@ -38,6 +39,9 @@ foreach my $incl (@possible_headers) {
 
 sub conf_via_ncurses_config {
    return () unless $found_header;
+   # NOTE the ncurses*-config executables are shell scripts so `sh` is
+   # required.
+   return () unless File::Which::which('sh');
 
    my @nc_configs = qw(
       ncurses6-config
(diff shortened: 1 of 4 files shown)


IO/Browser/Makefile.PL
```diff
@@ -20,29 +20,75 @@ $hash{'clean'}{FILES} .= ' browse$(OBJ_EXT) browse$(EXE_EXT) Browser.c Browser.p
...
+   return () unless $found_header;
+
+   my @nc_config = qw(
+      ncurses6-config
+      ncursesw6-config
+      ncurses5-config
+      ncursesw5-config
+   );
+   for my $nc_conf (@nc_config) {
+      local $ENV{NCURSES_CONFIG_PROG} = $nc_conf;
```
zmughal, line 42: Using an environment variable is absolutely necessary otherwise MSYS2 will think it can rewrite the path for you which causes it to hang (I think this is why).

mohawk2: Is there any way to use a list form? That `sh -c '...'` hurts my eyes ;-) Also, would it make sense to at least use File::Which first?

zmughal: I'd like `File::Which` if we can make that a configure dep (I think it is just runtime now). The only exception is that `File::Which` will not work on Windows because it only considers non-empty `PATHEXT` as executable (i.e., it has to have a suffix).

I could use the list form with `open( my $fh, "-|", ...)` even if that's a bit longer (but I'm happy to make that change). My preference is to use `Capture::Tiny`, but I didn't want to add another configure dep in this PR (even though both `File::Which` and `Capture::Tiny` are small).

mohawk2: I would be Ok with F:W as a configure dep. I'm with you on preferring not to add deps. And my preference is to use the `open` list construct you give, maybe with a tiny utility function if using several times? Is it possible that would bypass the `ENV` thing as well?

IO/Browser/Makefile.PL
```diff
@@ -20,29 +21,85 @@ $hash{'clean'}{FILES} .= ' browse$(OBJ_EXT) browse$(EXE_EXT) Browser.c Browser.p
...
-
-my $libstring;
-foreach my $libr ( qw( curses ncurses ncursesw ) ) {
-   if (eval { check_lib(lib=>$libr) }) {
-      $libstring = '-l' . $libr;
-      last;
+}
+
+sub conf_via_ncurses_config {
+   return () unless $found_header;
```
zmughal, line 41: Should probably use F::W on ‘sh’ here too.

mohawk2:
I've merged your `devops` PR so please remove the `ref:` in this one, and then merge as you see fit.

PR #423, opened 2023-04-11
Address clang compiler warnings in PDL 2.082

d-lamb:
These commits address most, but not all, of the clang compiler warnings that I see with PDL 2.082.  Most have to do with the newly supported datatypes such as (complex) Long Double, and avoids possible truncation-of-value problems.

### DIFF
--- a/Basic/Ops/ops.pd
+++ b/Basic/Ops/ops.pd
@@ -408,7 +408,7 @@ cfunc('re', 'creal', 1, 'Returns the real part of a complex number.',
 cfunc('im', 'cimag', 1, 'Returns the imaginary part of a complex number.',
   '$complexv() = creal($complexv()) + I * $b();'
 );
-cfunc('_cabs', 'cabs', 1, 'Returns the absolute (length) of a complex number.', undef,
+cfunc('_cabs', 'fabs', 1, 'Returns the absolute (length) of a complex number.', undef,
     PMFunc=>'',
 );
 my $rabs_code = '


mohawk2:
Thank you for this! I did see the notifications on the IRC channel you'd pushed to this branch, and already took a look.

Some of these changes are a really good idea, and I am about to cherry-pick them straight onto `master`, so I'll also rebase past those in a moment and force-push. I have comments about the others, but first things first!

Basic/Core/pdlcore.c
```diff
@@ -1203,5 +1203,5 @@ uint64_t pdl_pdl_seed() {
 	(void)time(&seconds);
 	/* End of Perl-specific symbols */
 	s = (uint64_t)seconds;
-	return abs(((s*181)*((pid-83)*359))%104729);
+	return ((s*181)*((pid-83)*359))%104729;
```
mohawk2, line 1206: I had to check but you (and clang) are quite correct, that unsigned arithmetic always wraps, so thank you. Pushing this to `master`.

mohawk2 (changes requested):
The one remaining change for this (making a complete copy of `cfunc` with one change) is slightly problematic, for two reasons:
- I'm not convinced it's done in the right way: C99 (which we require for exactly this, complex numbers) defines that `tgmath` supports `fabs` for complex numbers including `complex long double`: https://en.cppreference.com/w/c/numeric/tgmath
- Copy-paste programming creates problems for future maintainers; if this change *were* necessary, far better to do it by adding arguments to something, similar to the `NoTgmath` in `ufunc`

Could you try instead replacing this line:

`cfunc('_cabs', 'cabs', 1, 'Returns the absolute (length) of a complex number.', undef,`

so the second argument is `fabs`, and delete the `cfuncl` stuff, and see what clang says?

d-lamb:
Essentially undoing my original commit to ops.pd and making the change you suggested (`cabs` -> `fabs` in that one line) causes no additional clang warnings, and passes `make test`.  It's fine with me if you want to just drop the one remaining commit here on the floor and make that change directly in master.

mohawk2:
I instead adjusted your commit to do this (which had the additional benefit of leaving all the other `cfunc` users with `complex long double`), and have now merged it. Thank you!

PR #425, opened 2023-04-13
PDL demos: Add to the documentation how to write a demo

HaraldJoerg:
Here are some hints which might make it easier for first-time writers of PDL demos (as discussed on IRC)

### DIFF
--- a/Demos/Demos.pm
+++ b/Demos/Demos.pm
@@ -192,7 +192,7 @@ Each demo module must provide these class methods:
 
 =item info
 
-Return a two-element list: of strings: a single keyword (probably
+Return a two-element list of strings: a single keyword (probably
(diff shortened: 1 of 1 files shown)


Demos/Demos.pm
```diff
@@ -189,19 +192,20 @@ Each demo module must provide these class methods:
 
 =item info
 
-Return a two-element list: a single keyword (probably lower-case),
-and a short description of the demo.
+Return a two-element list: of strings: a single keyword (probably
```
mohawk2, line 195: A tiny nit: you say "list:", and the ":" is unnecessary. One could argue that the description should make it very clear they are strings, but I don't mind being more explicit.

mohawk2 (approved):
Thank you for this! I will delete the `:` as noted, and then merge it.

PR #426, opened 2023-04-13
PDL Demos:  simple demos to prepare sound (sine waves) with PDL

HaraldJoerg:
The demo uses PDL to calculate the samples of sine waves in one line, and then passes the "raw" PDL data to an external sound player.

I don't know whether this is good enough for a demo, but for me it was enlightening to see that raw sound can be calculated without looping over Perl arrays, and then directly passed to a sound player.

Since this is my first step with PDL, I'd welcome a review anyway.  I might easily have misunderstood things.

### DIFF
--- a/Demos/Makefile.PL
+++ b/Demos/Makefile.PL
@@ -15,6 +15,7 @@ WriteMakefile(
     'TriD2.pm'    => '$(INST_LIBDIR)/Demos/TriD2.pm',
     'TriDGallery.pm'    => '$(INST_LIBDIR)/Demos/TriDGallery.pm',
     'BAD_demo.pm'       => '$(INST_LIBDIR)/Demos/BAD_demo.pm',
+    'Sound.pm'          => '$(INST_LIBDIR)/Demos/Sound.pm',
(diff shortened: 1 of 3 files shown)


mohawk2:
Thank you for this! How do you feel about looking at Alien::Gnuplot and copying it to make an installer for a Linux/Windows/MacOS-capable sound player?

@zmughal What do you think? Do you have capacity to help Harald?

mohawk2:
@HaraldJoerg Thank you for your continuing work on this! Please could you not merge `master` into your branch, but instead switch to your `master`, `pull` that, then `rebase` your demo branch against that, then force-push?

HaraldJoerg:
Ouch... merging the master branch was not intended at all...  My bad handling of the GitHub is to blame.

About creating an Alien module for a player: I feel that the added value is too small if it is just for running a demo.  I am still unsure where my experiments with sound will [float to](https://xkcd.com/1/).  PDL ndarrays are not too far away from Microsoft WAV files which can be played on any platform because they contain the necessary metadata...

PR #427, opened 2023-04-14
More docs for PDL demo creators:  Bookkeeping tasks

HaraldJoerg:
Mention that new files need to be added to Makefile.PL and MANIFEST,
otherwise the CI actions won't succeed.


mohawk2 (changes requested):
I feel like describing this in a more abstract way is better here: instead, please recommend that people ensure the demo module is included in a distribution, then mention the `MANIFEST` etc as an example of that.

Issue #430, opened 2023-04-23
append can give wrong shape; dimensions of size 0 are involved


falsifian:
```
pdl> p append(zeroes(2,0), zeroes(3,0))->shape
[3 0]
```

I'm pretty sure that should be `[5 0]`.

mohawk2:
As am I!

Issue #429, opened 2023-04-23
`zeroes(1,1,0) & zeroes(1,1,0)` segfaults


falsifian:
```
pdl> zeroes(1,1,0) & zeroes(1,1,0)
Segmentation fault (core dumped) 
```

This is PDL 2.081 on OpenBSD current on amd64.

Issue #428, opened 2023-04-23
whereND drops all dimensions when mask is all zeroes


falsifian:
One of these is not like the others:

```
pdl> p zeroes( 2, 3, 1 )->whereND( pdl '1 1' )->shape
[2 3 1]
pdl> p zeroes( 2, 3, 1 )->whereND( pdl '1 0' )->shape
[1 3 1]
pdl> p zeroes( 2, 3, 1 )->whereND( pdl '0 1' )->shape
[1 3 1]
pdl> p zeroes( 2, 3, 1 )->whereND( pdl '0 0' )->shape
[0]
```

I would expect the last output to be `[0 3 1]`.

falsifian:
This is PDL 2.081 on OpenBSD current on amd64.

Here's another behaviour that seems like a bug:

```
pdl> p zeroes( 2, 1 )->whereND( pdl '1 1' )

[
 [0 0]
]

pdl> p zeroes( 2, 0 )->whereND( pdl '1 1' )
Stringizing problem: Error in splitdim:nsp 2 cannot be greater than dim 0.     eval {...} called at /home/falsifian/perl5/lib/perl5/amd64-openbsd/PDL/Core.pm line 2849
        PDL::string(PDL=SCALAR(0x15b28c54b20), undef, "") called at /home/falsifian/perl5/bin/perldl line 409
        main::p(PDL=SCALAR(0x15b28c54b20)) called at (eval 343) line 4
        main::__ANON__() called at /home/falsifian/perl5/bin/perldl line 682
        eval {...} called at /home/falsifian/perl5/bin/perldl line 682
        main::eval_and_report("p zeroes( 2, 0 )->whereND( pdl '1 1' )\x{a}") called at /home/falsifian/perl5/bin/perldl line 618
        main::process_input() called at /home/falsifian/perl5/bin/perldl line 638
        eval {...} called at /home/falsifian/perl5/bin/perldl line 638
 at /home/falsifian/perl5/lib/perl5/amd64-openbsd/PDL/Core.pm line 2880, <STDIN> line 50.
        PDL::string(PDL=SCALAR(0x15b28c54b20), undef, "") called at /home/falsifian/perl5/bin/perldl line 409
        main::p(PDL=SCALAR(0x15b28c54b20)) called at (eval 343) line 4
        main::__ANON__() called at /home/falsifian/perl5/bin/perldl line 682
        eval {...} called at /home/falsifian/perl5/bin/perldl line 682
[...]

PR #431, opened 2023-04-25
Test some operations with empty ndarrays.

falsifian:
The new tests all fail, two of them by segfaulting.

This covers GitHub issues #428, #429 and #430, and also an unreported issue that random(1,1,0) segfaults.


falsifian:
This is WIP: the issues should probably be fixed before this is merged.

Issue #439, opened 2023-06-17
Fails to build with HDF4 4.2.16 (undefined symbol: HDfreespace)


sebastic:
HDF4 4.2.16 removed aliases in `hdfi.h`.

4.2.15 had:
```c
/************************************************************************** 
*  Macros to work around ANSI C portability problems.
**************************************************************************/ 
#ifdef DUMBCC
#define CONSTR(v,s) char *v=s
#else   
#define CONSTR(v,s) static const char v[]=s                                 
#endif 

/* Old-style memory allocation function aliases -QAK */                     
#define HDgetspace HDmalloc                                                 
#define HDclearspace HDcalloc 
#define HDregetspace HDrealloc                                              
#define HDfreespace HDfree
```
These are removed in 4.2.16, and notes that the `HD*` aliases should not be used:
```c
/************************************************************************** 
 *  Memory functions
 **************************************************************************/

/* DO NOT USE THESE MACROS */

/* These will be removed from a future version of the library and are       
 * only kept here to avoid breakage in programs that unwisely used          
 * them.    
 */     
#define HDmalloc(s)     malloc(s) 
#define HDcalloc(a, b)  calloc(a, b)                                        
#define HDfree(p)       free(p)
#define HDrealloc(p, s) realloc(p, s)                                       

/* Macro to free space and clear pointer to NULL */                         
#define HDfreenclear(p)                                                                                      \
    {                                                                                                        \
        free(p);                                                                                             \
        (p) = NULL;                                                                                          \
    }   
```

PR #440, opened 2023-06-17
Use malloc/free instead of HDgetspace/HDfreespace.

sebastic:
As noted in #439, hdf4 4.2.16 removed deprecated aliases which were still used by PDL.

Fixes: #439


mohawk2:
Thank you!

PR #443, opened 2023-07-16
Fix build on FreeBSD by allowing "amd64"

sunpoet:
It also unifies the pattern in various places.


mohawk2:
Thanks!

mohawk2:
For another time, it will probably help if you use a different branch locally rather than `master` - see the FAQ for tips.

PR #444, opened 2023-07-20
Fix builds on MacOS Ventura

deriamis:
Apple has apparently changed the location of glut.h in recent versions of MacOS. This is a small change to get things building again on the platform.

### DIFF
--- a/Changes
+++ b/Changes
@@ -7,6 +7,8 @@
 - fix GIS::Proj to work better on Windows (#456) - thanks @shawnlaffan
 - fix conversion of ulonglong to Perl scalar and therefore stringification (#458) - thanks @vadim-160102 for report
 - fix modulo for ulonglong and on Windows
+- update Opt::Simplex docs (#452) - thanks @KJ7LNW
+- fix Graphics::OpenGLQ on MacOS Ventura (#452) - thanks @deriamis
 
 2.084 2023-05-21
 - reduce size of PDL_KLUDGE_COPY_X macro to <4096 in line with C standard, to fix for older clang on many BSD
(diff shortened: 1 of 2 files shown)


mohawk2:
Since the location has only changed on MacOS, you'll need to use a `#if` or similar to keep it working elsewhere, not just change it globally.

deriamis:
> Since the location has only changed on MacOS, you'll need to use a `#if` or similar to keep it working elsewhere, not just change it globally.

The change is already wrapped in an `#ifdef HAVE_AGL_GLUT`. My understanding is that AGL is the Apple-specific OpenGL implementation. Is this not the case?

mohawk2:
Ah, excellent point! Is there a version of MacOS where this changed, which could be added as a comment and/or `#if`-ed as well?

deriamis:
> Ah, excellent point! Is there a version of MacOS where this changed, which could be added as a comment and/or `#if`-ed as well?

I don't know at the moment, but I can try to find out. I'll let you know what I discover.

However, that brings up the question of what we're trying to support for builds. It may be the case that the change happened a while ago in a version Apple no longer supports. Do we still need the `#if` if that's the case?

mohawk2:
What I'm after is adding at least a comment in the `.h` file showing what version that new location applies from, if possible. I'm not overly concerned with support vastly far back in every possible combination of OS, version, library, etc.

deriamis:
@mohawk2 From what I'm seeing, the  `GLUT/glut.h` include has actually been the documented (by Apple) way to use GLUT since at least 10.4, if not before. I see example code from 2005 that used it that way. 

However, it looks like there was a symlink that made `OpenGL/glut.h` work right up until I think Sierra, if not maybe Mojave (when Apple deprecated OpenGL). In any case, this change is likely backwards-compatible farther back than we would ever need to care about.

mohawk2:
Thank you! This will be released soon.

PR #449, opened 2023-07-25
allow for nested coretests and regular tests in t/

djerius:


zmughal:
This looks good. I see that this PR is separate from creating the nested tests?

djerius:
> This looks good. I see that this PR is separate from creating the nested tests?

Yup.  I didn't want to generate too much churn if this needed tweaking.

PR #452, opened 2023-08-21
doc: update SEE ALSO in PDL::Opt::Simplex

KJ7LNW:
Add a link to the out-of-tree PDL::Opt::ParticleSwarm optimizer.

Also update the broken link from 1998 to chem1.nrl.navy.mil with something that "works" from archive.org


mohawk2:
Thanks! I tweaked it a bit since I didn't think it needs an explicit number on `=over`.

KJ7LNW:
Awesome, have a great week!

PR #453, opened 2023-08-23
Reorganize and modlarize tests in t/primitive.t

djerius:
This commit reorganizes and modularizes t/primitive.t, moving related tests into separate files.
It switches over to Test2::V0.

Splitting the large monolithic t/primitive.t into smaller chunks will speed up tests if parallel testing (e.g. prove -j or yath -j) harnesses are used.


zmughal (approved):
Looks good to me. And clean!

While it does add a dependency on `Test2::V0`, I believe that it is widely used enough to justify using it (I use it myself). Plus modern `Test-Simple` uses `Test2` anyway.

djerius:
> Looks good to me. And clean!
> 
> While it does add a dependency on `Test2::V0`, I believe that it is widely used enough to justify using it (I use it myself). Plus modern `Test-Simple` uses `Test2` anyway.

Thanks for the review.

zmughal:
Just to add this this, `Test2::V0` will be a dual life module in Perl 5's core in the next release <https://github.com/Perl/perl5/pull/21481>, [perl5394delta](https://github.com/Perl/perl5/blob/blead/pod/perl5394delta.pod).

Issue #441, opened 2023-07-03
gis_proj.t fails on Windows


shawnlaffan:
The gis_proj.t tests are failing under windows.  The reason is that the Alien::proj bin dirs are not in the path when the DLL is loaded.  (Edit - when Alien::proj is detected by the ```perl Makefile.PL``` call).

Adding this code to the test file gets things to pass but obviously won't fix later uses of PDL::GIS::Proj.  It is easy enough to create a variant that does not depend on the Env module.

```perl
BEGIN {
    if ($^O =~ /MSWin32/) {
        use Env qw '@PATH';
        unshift @PATH, Alien::proj->bin_dirs;
    }
}
```

Adding the above to code to ```Proj.pd``` would be a more general solution but the ```add_pm``` sections seem to be insrted into the pm file too late to have the desired effect.

If I add the equivalent code to Alien::proj then that code can be simplified.  PDL::GIS::Proj will still need to load Alien::proj, which it currently does not do.

mohawk2:
I believe that conceptually, Alien modules are supposed to load the library and make it available, so `use`ing it should do that. This would mean modifying that library to do what you said above (though let's not add a dep on Env!), then `use`ing it in PDL::GIS::Proj.

Do you have capacity to make those changes and PR them?

mohawk2:
Sorry for the long wait. Merged, will release shortly.

PR #456, opened 2023-09-24
GIS::Proj - Use Alien::proj

shawnlaffan:
And add its bin dirs to the path on Windows
so it can function.

Done using pp_addbegin as pp_addpm happens
after the dynaload call, which is too late.

Updates #441


shawnlaffan:
Would be good to get this reviewed and/or committed. 

If the build detects Alien::proj and it then unconditionally adds the proj components.  These then fail, which means one cannot install or update PDL on windows if Alien::proj is already installed (unless the tests are ignored or skipped, of course).

PR #472, opened 2024-04-11
Modified documentation of srand to make it clear it is not

chrisarg:
exported and does not overwrite Perl's builtin srand.

### DIFF
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2179,8 +2179,18 @@ As of 2.062, the generator changed from Perl's generator to xoshiro256++
 
 =for usage
 
+From within pdl:
+
  srand(); # uses current time
  srand(5); # fixed number e.g. for testing
+
+Since srand is a class function and is not exported, it does not ovewrite
+Perl's builtin srand. Hence the code above will not seed PDL's generator
+when invoked from within a Perl application. Being a class method means 
+that PDL->srand() will generate an error too. Do this instead:
+
+ PDL::srand::srand(); # uses current time
+ PDL::srand::srand(5); # fixed number e.g. for testing
 
 EOF
 	PMCode=>pp_line_numbers(__LINE__, <<'EOD'),
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2191,6 +2191,33 @@ that PDL->srand() will generate an error too. Do this instead:
 
  PDL::srand::srand(); # uses current time
  PDL::srand::srand(5); # fixed number e.g. for testing
+ 
+See why keeping the srand's straight is important:
+
+ srand(3);
+ say PDL->random(3);
+ [0.0675697331360393 0.599075396409301 0.857557628839386] # output PDL1
+ 
+ print "(";print "$_" for map { rand() } 1 .. 3 ; print ")\n";
+ (0.783234962103055 0.86367337321596 0.31170834325361)      # output Perl1
+
+
+ srand(3);
+ say PDL->random(5); 
+ [0.924203795779931 0.333781400627343 0.831579265248175] # != output PDL1
+
+ print "(";print "$_" for map { rand() } 1 .. 3 ; print ")\n";
+ (0.783234962103055 0.86367337321596 0.31170834325361) # same as output Perl1
+ 
+Contrast to 
+
+ PDL::srand(3);
+ say PDL->random(3);
+ [0.18631707882894 0.0349571597540506 0.614691270370328]
+ 
+ PDL::srand(3);
+ say PDL->random(3);
+ [0.18631707882894 0.0349571597540506 0.614691270370328] # reproduces!
 
 EOF
 	PMCode=>pp_line_numbers(__LINE__, <<'EOD'),
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2184,15 +2184,14 @@ From within pdl:
  srand(); # uses current time
  srand(5); # fixed number e.g. for testing
 
-Since srand is a class function and is not exported, it does not ovewrite
-Perl's builtin srand. Hence the code above will not seed PDL's generator
-when invoked from within a Perl application. Being a class method means 
-that PDL->srand() will generate an error too. Do this instead:
+This  code above will not seed PDL's generator, but Perl's rand in Perl.  
+So if you want to seed PDL's generator, you will have to do something like this
+in the Perl interpreter:
 
- PDL::srand::srand(); # uses current time
- PDL::srand::srand(5); # fixed number e.g. for testing
+ PDL::srand(); # uses current time
+ PDL::srand(5); # fixed number e.g. for testing
  
-See why keeping the srand's straight is important:
+See why keeping the srands straight is important (try those within Perl):
 
  srand(3);
  say PDL->random(3);
--- a/Basic/Primitive/primitive.pd
+++ b/Basic/Primitive/primitive.pd
@@ -2179,14 +2179,12 @@ As of 2.062, the generator changed from Perl's generator to xoshiro256++
 
 =for usage
 
-From within pdl:
+From within C<perldl>:
 
  srand(); # uses current time
  srand(5); # fixed number e.g. for testing
 
-This  code above will not seed PDL's generator, but Perl's rand in Perl.  
-So if you want to seed PDL's generator, you will have to do something like this
-in the Perl interpreter:
+But, from within the Perl interpreter, you will have to do something like this:
 
  PDL::srand(); # uses current time
  PDL::srand(5); # fixed number e.g. for testing


mohawk2:
Thank you for your first pull request! Currently you're leaving some code you then say doesn't work.

Please could you instead improve this further, so that you change the code to something that works? Nothing in the documentation is sacred, and everything can be improved. Mostly that happens through people with fresh eyes seeing the mistakes, and changing them :-) Probably something like a call to `PDL::srand` with a fixed value, then a call to `random` printing out the value it makes?

Basic/Primitive/primitive.pd
```diff
@@ -2179,9 +2179,45 @@ As of 2.062, the generator changed from Perl's generator to xoshiro256++
 
 =for usage
 
+From within pdl:
```
mohawk2, line 2182: Please change this to say "within `C<perldl>`", because "within pdl" is very unclear :-)

Basic/Primitive/primitive.pd
```diff
@@ -2179,9 +2179,45 @@ As of 2.062, the generator changed from Perl's generator to xoshiro256++
...
+
  srand(); # uses current time
  srand(5); # fixed number e.g. for testing
 
+This  code above will not seed PDL's generator, but Perl's rand in Perl.
```
mohawk2, line 2187: I asked you to remove the code that doesn't work. Changing it to code that *does* work, and adding why it needs to be that, is fine. But you're leaving it with the first thing there as something that doesn't work, which is very unhelpful.

Please change this. It won't get merged until that's done.

mohawk2 (changes requested):

chrisarg:
I thought I pushed them out back for your review?? Let me take a look.

mohawk2:
Please do. Please also review this:

> From within C<perldl>:
>
>   srand(); # uses current time
>   srand(5); # fixed number e.g. for testing
>
> But, from within the Perl interpreter, you will have to do something like this:

What is it you think is the difference between `perldl` and "the Perl interpreter"? As written, it doesn't make sense.

Additionally, and quite importantly, `srand` in `perldl` (or any Perl script that does `use PDL`), does in fact set the PDL random seed, such that doing `srand(5); p +PDL->random(3)` gives the same result over and over. In other words, it **is** exported, and therefore **does** overlay Perl's `srand` in the local namespace.

sisyphus:
One strange thing I did notice (while trying to understand what this issue is about) is that the following one-liner produces different output each time I run it:
```
C:\>perl -MPDL -E "srand(3); say rand();"
0.422170424674736

C:\>perl -MPDL -E "srand(3); say rand();"
0.000277820349769087
```
Is that intended and documented and expected behaviour ?
In order to experience the repeating behaviour that I expect, I must <b>not</b> load PDL:
```
C:\>perl -E "srand(3); say rand();"
0.783234962103055

C:\>perl -E "srand(3); say rand();"
0.783234962103055
```
Apologies if this observation is not relevant to this issue.

mohawk2:
@sisyphus If you `use PDL`, that imports (among many, many other things) `srand` from PDL::Primitive. Calling that has no effect at all on Perl's random-number generator. Therefore, calling it, then calling Perl's `rand()`, will be as though you didn't call Perl's `srand`, because you didn't, and you will get, uh, random-ish numbers.

marioroy:
This is somewhat sad news for folks using PDL with MCE or MCE::Shared. The following code no longer outputs the same random numbers.

```perl
use v5.030;
use PDL;
use MCE;

srand(3);

MCE->new(
  max_workers => 4, user_func => sub {
    MCE->print(MCE->wid, " ", rand(), "\n");
  }
)->run;
```

```text
$ perl ex.pl | sort
1 0.455586975281225
2 0.224720416431413
3 0.993853857581602
4 0.76298729873179

$ perl ex.pl | sort
1 0.778804464106546
2 0.547937905256735
3 0.317071346406923
4 0.0862047875571115

$ perl ex.pl | sort
1 0.49638699840671
2 0.265520439556898
3 0.0346538807070864
4 0.803787321857275
```

Thank you, @sisyphus for the PerlMonks [thread](https://perlmonks.org/?node_id=11159773). I will attempt a workaround for MCE and mindful of `PDL` and `PDL::Lite`.

```perl
# Call PDL's random() function if exported i.e. use PDL.
$self->{_seed} = $INC{'PDL/Primitive.pm'} && eval "$caller->can('random')"
    ? int(PDL::Primitive::random() * 1e9)
    : int(CORE::rand() * 1e9);
```

Is there a sane way to know if srand was exported by PDL?

chrisarg:
This is what I was trying to get at. I will write a more extensive response
to you later in the weekend (i thought i had put some of that in the pull,
but maybe i did it wrong).  I am trying to wrap up the repos for the perl
talk (bunch of PDL stuff there too). I also have to check whether there was
a recent shift in behavior of srand in different versions of the PDL.

It seems that when you overwrite the srand , the native perl rand no longer
can be seeded.

chrisarg:
I looked into this, because I was intrigued

Code for all scenarios is
###########################################################################


use PDL;
use v5.38;
srand(3);

say "PDL version : ", $PDL::VERSION;
 srand(3);
 say "PDL : " , PDL->random(3);
 print "Perl : (";print "$_" for map { rand() } 1 .. 3 ; print ")\n";

 srand(3);
 say "PDL : ",PDL->random(3);
 print "Perl : (";print "$_" for map { rand() } 1 .. 3 ; print ")\n";
############################################################################

For example, under *PDL* *2.057 *both PDL's random and the builtin rand
produce consistent results

PDL version : 2.057
PDL : [0.78323496 0.86367337 0.31170834]
Perl : (0.2670276363814490.5418611971081940.433702219495885)
PDL : [0.78323496 0.86367337 0.31170834]
Perl : (0.2670276363814490.5418611971081940.433702219495885)

Under *PDL 2.084 *(also under *2.067, 2.072, 2.089*) srand does not set
bultin-perl's seed
PDL version : 2.084
PDL : [0.18631708 0.03495716 0.61469127]
Perl : (0.8141075503817810.9808434055625810.883542824807716)
PDL : [0.18631708 0.03495716 0.61469127]
Perl : (0.08131618387682950.3513600075503580.658677388433702)


*BUT change use PDL; to use PDL::Lite; and the code produces*

PDL version : 2.067
PDL : [0.19136166 0.44822994 0.27954422]
Perl : (0.7832349621030550.863673373215960.31170834325361)
PDL : [0.59997668 0.59428229 0.69931671]
Perl : (0.7832349621030550.863673373215960.31170834325361)


To get back to the issue :* my suggested edit to the help is obviously
wrong if one imports PDL, but it is right if one imports PDL::Lite *.

There are two somewhat interrelated issues here:

1) whether it makes sense to overwrite Perl's srand via PDL's srand to
begin with (IMHO this is a no, and I will explain why below)

2) the inconsistency in the output of Perl scripts that seed and then use
PDL to generate random numbers  depending on the importation of PDL::Lite
v.s. PDL

My personal opinion about #1 is that it creates an issue for people who for
whatever reason may want to generate random numbers using either PDL's
random and he builtin rand.

The inconsistent overwrite also creates the issue with #2, which is a major
headache for people who don't want the overhead of importing the entire PDL
in their scripts (e.g. if they are forking stuff, or if the startup time of
the script is critical) and just use the Lite version.

Looking at how other software systems/languages manage RNGs:
1) R is a good example of an interpreted language that features literally
dozens of scripts that come with their own RNG. Those don't overwrite the
builtin RNG, and just settle using a different name for the seed function
of their own rng
(e.g. look at the documentation of dqrng that also provides access to the
Xoshiro RNGs) https://cran.r-project.org/web/packages/dqrng/dqrng.pdf

2) others encapsulate their RNG into objects (e.g. this is how the GSL does
this, and so does Intel's oneAPI) to avoid conflicting with whatever
builtin exists (and also to provide opportunities for thread safety to the
RNG).

Ultimately, the decision is up to the PDLPorters on how to handle this, but
I feel that at least warning that one should not expect srand to seed the
PDL random number generator if the use PDL::Lite is reasonable at this
point.
Perhaps one could also simultaneously seed Perl's builtin RNG when the
override is set up (this could be a good use of Hook::LexWrap). I am sure
there are many other designs that work as well, (or likely better) than the
two I sugested in the last 2 lines

mohawk2:
The situation has moved a little bit, which is why this page now says there is a "conflict that must be resolved". PDL's RNG-seeder is now called `srandom`, thereby not clashing with Perl's. The docs now say when these things changed.

Please update this PR to reflect the current reality. Or, it's actually possible those changes render this now moot?

chrisarg:
Aha!
If it is called srandom, then the point may be avoided if one uses srandom
consistently . However,  from the code above it seems that srand can still
seed the PDL generator if one "use PDL;", but not if one "use PDL::Lite'"
(at least up to 2.089). This can still give surprises e..g to those who are
not aware of the nuance.
Perhaps, a note in the documentation, and asking people to consider moving
to srandom, because of the inconsistency in the overwriting of the function
srand.

Christos

mohawk2:
If you don't use `srandom` as of the current (dev-)released version of PDL, 2.089_01, you won't affect PDL's RNG at all. Consistency has nothing to do with it, nor does which of the loading modules is chosen. `srand` has gone back to affecting strictly the Perl RNG. There is no point in documenting past versions of PDL beyond what has already been recorded in the `srandom` doc.

chrisarg:
Thanks for the clarification. Then there is no point in discussing PR #472
any further, as the issues seems to have been resolved. Looking forward to
seeing PDL 2.089_1 make it past the dev_release stage!

PR #481, opened 2024-06-07
A contribution to developer docs

HaraldJoerg:
Here are two suggestions for documentation to help people getting started with PDL development.  I'll probably add more while I get deeper into the code.


mohawk2:
Thanks, this is all good stuff! Would you like me to merge it as-is, or do you feel like making `DEVELOPMENT` actually be a proper POD document in `Basic/Pod`, e.g. PDL::DeveloperGuide?

HaraldJoerg:
Oooh... converting it to POD would make a lot of sense IMHO.  It would be discoverable from MetaCPAN and readable with a browser, enjoying a table of contents, and links to other places... Maybe some synchronization with https://github.com/PDLPorters/pdl/wiki/Developer%27s-Corner would not hurt either.  I'll do that.

HaraldJoerg:
Ah, and there's also https://pdl.perl.org/?page=develop (from pdlporters.github.com).  Maybe the wiki or GitHub pages would be the best place to present information, and reduce the DEVELOPMENT file to a pointer?  Three different places cause synchronization work without much benefit

mohawk2:
I agree with the idea of "only one place". I don't know if the wiki is very useful; I'd suggest we suck all the useful info from it into the website/dev docs and turn it into a pointer.

That said, I'm open to more specific ideas!

HaraldJoerg:
Ok, let's be more specific: Let me first clarify the issue: The DEVELOP file is not that easy to discover. It is mentioned in README.md, but MetaCPAN does not show that information, and on GitHub one must scroll past two screenfuls of directories and files before the Readme even starts. 

For MetaCPAN discoverability I suggest to create a file Contributing.pod: This file will be picked up by MetaCPAN and linked under the "How to Contribute" navigation item (as of now, not many CPAN modules make use of this).

The contents of the DEVELOP file and the github.io page https://pdl.perl.org/?page=develop should be merged and updated.  The DEVELOP file should be deleted, and README.md adjusted accordingly.

As for the location of the developer documentation: This could be either on pdl.perl.org, i.e. written in HTML, or in the repository under Contributing.pod.  Both formats, POD and HTML, are good enough and better suited for guidelines as plain text.  I have a slight preference for Contributing.pod, because then it is in the same repository as PDL itself.

In any case, the PDL developer page should keep the short content it has and link to detailed developer documentation.

mohawk2:
Please can you rebase your branch against current `master`? Merge commits, as you've done here, are not acceptable due to the confusion they create; for instance, your PR now appears to try to make a change in `Core.xs` that I made recently.

You'll see your PR now fails an author test; you've added a file, but not added it to the `MANIFEST`. That needs fixing. Also, you've ignored my suggested name for the developer guide. Please call it what I suggested - this is partly because "contributing" is usually a name for either community guidelines, or specific guidance on making pull requests, which this will be more than. I also note that you're still leaving, and indeed adding to, the `DEVELOPMENT` file. Instead that needs incorporating into this new PDL::DeveloperGuide.

The reason that location is better for developer type docs is that when changes get made to PDL and PRed, the dev guide can be reviewed alongside it. That makes it a better store for that information than either a wiki or the website, neither of which really get maintained. Removing information from the wiki and the website into one "source of truth" that *is* maintained is a far better approach.

HaraldJoerg:
Sorry, github handling problem here... I updated my fork's master branch using the web interface which is a bad idea.  I hope I managed to clean up the history.

Not deleting the DEVELOPMENT file in that commit was intentional, to allow easier side-by-side review of the contents of both files - once the content is approved, the deletion is as easy as another commit.

Regarding the name of the POD document:  I am fine with publishing it as PDL::DeveloperGuide, I'll make a commit for renaming that.  Still I think that having a Contributing.pod (or .md) makes sense: This is where MetaCPAN will pick it up: MetaCPAN is _the_ most important place to discover Perl stuff. Yet another file in the distribution is difficult to discover: https://metacpan.org/dist/PDL is a looooong list with a seemingly random distribution of stuff between "Documentation" and "Modules".  A short Contributing.pod pointing to the PDL::DeveloperGuide can go a long way.

mohawk2:
Why do you think MetaCPAN will pick up the "contributing" file? MetaCPAN talks about a file located at the top with content similar to what I described above. It won't really add any value, and that's not where people would look for a developer guide. It will be fine to link to such from both the main PDL docs, and the PDL "quickstart", but this "contributing" idea is not helpful.

The distribution page for PDL lists lots of modules because PDL has lots of modules. There's not anything "random" about it.

HaraldJoerg:
History has been rewritten to squash the two commits on the way to DeveloperGuide.pod into one.

mohawk2:
Thanks, this is great!

Some notes for the next brave soul who wants to update the shiny new developer guide:

- The "PDL config" stuff is largely irrelevant now, since that was never a good way to do configuration but was necessary until the conditionally-runnable tests got moved to where they made sense, next to the conditionally-built code.
- The bit about `Makefile.PL` needing to not do `use PDL::Core::Dev` isn't really true, but doesn't do enormous harm.
- Also, it's unlikely there will be any further "ingesting" of distributions, so that bit can decently go.
- This doc now needs linking from the main PDL doc, and also from the "QuickStart" one.
- Somewhere in these POD docs the various notes in open "developer notes" issues want incorporating.
- There are the various

PR #482, opened 2024-06-09
scantree.pl: indent pod template fragments

haarg:
Pod in a heredoc is still treated as Pod by a pod viewer. This confuses any tool trying to look for all Pod files.

Indenting the Pod fragments means they won't be interpreted as Pod, and this indentation can be stripped off when writing the files.


mohawk2:
Thank you!

Issue #393, opened 2022-04-22
better document debug macros


fantasma13:
As discussed with Ed, please add using debugging macros to the documents.

mohawk2:
This needs to be in an updated "developer's guide":

To generate a stack-trace to help us debug a problem, please first of all do `make realclean`, ensure you have the latest released PDL, then if it is still happening, please follow this process:
- Rerun `perl Makefile.PL` (this will update the timestamps on `Makefile`s, prompting a full rebuild)
- Edit your generated Makefile to add `-g` to the `OPTIMIZE =` line to generate debug symbols
- Run: `make`
- Run: `gdb perl -ex 'run -Mblib t/tracking.t'`
- In gdb, run: `bt`
- Copy-paste the backtrace output which should show where the segfault originated

Then please open an issue on whichever repo is relevant (possibly PDL, possibly in this context PDL-OpenCV) with your `perl -V` output and PDL version, together with the stack-trace created above. When you do paste such output into a GitHub issue, please surround such blocks with ``` on their own lines. You can check out it will render with the "Preview" button above the issue-editing box.

mohawk2:
Techniques to be discussed should include the new `PDL_IF_BAD`, and recommend starting any design by thinking about complex values first, together with how to handle badvalues. An example of creative use of badvalues is mentioned in https://perlmonks.org/?node_id=11143676

Also to be discussed: code generation; cf PDL::Slatec, PDL::LinearAlgebra::Complex, and now PDL::OpenCV.

mohawk2:
Also in design thinking: broadcasting from the start (and pthreading, where relevant).

mohawk2:
Include techniques shown in #457

mohawk2:
Include techniques of `Example/address-pseudonymise` including `-x` mode to help spot old vs new behaviour.

mohawk2:
The fancy `loop()` and `OtherPars` stuff in https://www.perlmonks.org/?node_id=11157717 should probably be captured.

mohawk2:
The discussion of internals and dataflow in https://github.com/moocow-the-bovine/PDL-CCS/pull/5 should be extracted from.

mohawk2:
See #463 for notes on `doc-pp`.

mohawk2:
#451 has performance-analysing stuff, `cachegrind` in particular. #421 discusses other performance/implementation stuff.

mohawk2:
[...]

PR #486, opened 2024-06-27
More stuff for the DeveloperGuide, as suggested in GitHub#481

HaraldJoerg:
* Basic/Pod/DeveloperGuide.pod: Delete the sections "Access to PDL's configuration" and "Ingesting an external PDL module". Document $::PP_VERBOSE and $PDL::Graphics::TriD::verbose. Add a section "Generating a Stack Trace" taken from GitHub issue #393. Add some more mappings to the package->source file table.

* Basic/PDL.pm: Link to the Developer Guide under LANGUAGE DOCUMENTATION.

* Basic/Pod/FAQ.pod: Link to the Developer Guide in Q: 3.5.

* Basic/Pod/QuickStart.pod: Link to the Developer Guide. Also remove a phrase "until recently" which was recent ... in 1998, and don't mention HTML copies of the documentation which are no longer generated (and https://perldoc.perl.org exists).

### DIFF
--- a/Basic/Pod/DeveloperGuide.pod
+++ b/Basic/Pod/DeveloperGuide.pod
@@ -57,6 +57,74 @@ This is detailed in section 4.11 of L<PDL::FAQ>.
 
 =head1 PDL Developer Notes
 
+=head2 Getting (some) Debug Information
+
+PDL has some global variables which control diagnostic output:
+
+=over
+
+=item C<$::PP_VERBOSE>
+
+If set to a true value, L<PDL::PP> will print details while building
+the module processing C<pp_def> declarations.
+
+=item C<$PDL::Graphics::TriD::verbose>
+
+If set to a true value, the routines in L<PDL::Graphics::TriD> will
+report their progress.
+
+=item C<$PDL::verbose>
+
+Several modules produce diagnostic output to STDOUT if
+C<$PDL::verbose> has a true value.
+
+=item C<$PDL::debug>
+
+Several modules produce diagnostic output if C<$PDL::verbose> has a
+true value.
+
+=back
+
+
+=head2 Generating a Stack Trace
+
+To generate a stack-trace to help us debug a problem, please first of
+all do C<make realclean>, ensure you have the latest released PDL.
+Then, if it is still happening, please follow this process:
+
+=over
+
+=item Rerun perl Makefile.PL
+
+This will update the timestamps on Makefiles, prompting a full rebuild.
+
+=item Edit your generated top-level Makefile
+
+Add C<-g> to the C<OPTIMIZE => line to generate debug symbols (this gets
+propagated to subdirectories)
+
+=item Run: C<make>
+
+This builds PDL into the C<blib> directory.
+
+=item Run: C<gdb perl -ex 'run -Mblib t/tracking.t'>
+
+In gdb, run: C<bt>
+
+=item Copy-paste the backtrace output
+
+This should show where the segfault originated.
+
+=back
+
+Then please open an issue on whichever repo is relevant with your
+C<perl -V> output and PDL version, together with the stack-trace
+created above. When you do paste such output into a GitHub issue,
+please surround such blocks with ``` on their own lines. You can check
+out how it will render with the "Preview" button above the
+issue-editing box.
+
+
 =head2 Random Musings
 
 A (small) collection of random musings to note if you feel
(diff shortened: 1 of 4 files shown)


Basic/Pod/DeveloperGuide.pod
```diff
@@ -57,117 +57,79 @@ This is detailed in section 4.11 of L<PDL::FAQ>.
...
+=item C<$::PP_VERBOSE>
 
-git supports file-by-file commits so it is helpful
-to commit your changes to git a little at a time where
-each commit corresponds to a single change.  This makes
-it easy in the log to determine what was done and to
-locate any desired commit in case of issues that need
-to be resolved.
+If set to a true value, L<PDL::PP> will print details while processing
+C<pp_def>.
```
mohawk2, line 70: This is not really while "running" but while building. Those are both things that a developer is going to care about, but I'd like to be as accurate as possible while we're updating these docs :-)

mohawk2: There's also a `$PDL::verbose` which is worth mentioning.

Basic/Pod/DeveloperGuide.pod
```diff
@@ -57,117 +57,79 @@ This is detailed in section 4.11 of L<PDL::FAQ>.
...
 
-If your module requires external libraries or header files,
-add a section to perldl.conf.  The hash values with be available
-in your module's 'Makefile.PL' as $PDL::Config{WITH_MYMODULE},...
+git supports file-by-file commits so it is helpful
+to commit your changes to git a little at a time where
+each commit corresponds to a single change.  This makes
+it easy in the log to determine what was done and to
+locate any desired commit in case of issues that need
+to be resolved.
```
mohawk2, line 132: This isn't quite how things are best done - it looks like you're telling people to make one commit per *file*, which isn't right - it's one per "small, meaningful change (ideally each with an update to tests)". The reference to "file-by-file commits" needs removing and simplifying.

Basic/Pod/DeveloperGuide.pod
```diff
@@ -178,34 +140,39 @@ a PDL installation.  For some modules the .pm files are generated
...
+  | PDL::IO::Pic                           | IO/Pnm/pnm.pd                   |
+  | PDL::IO::Storable                      | IO/Storable/storable.pd         |
+  | PDL::Lvalue                            | Basic/Lvalue.pm                 |
+  | PDL::Math                              | Basic/Math/math.pd              |
+  | PDL::MatrixOps                         | Basic/MatrixOps/matrixops.pd    |
+  | PDL::NiceSlice                         | Basic/SourceFilter/NiceSlice.pm |
+  | PDL::Ops                               | Basic/Ops/ops.pd                |
+  | PDL::Primitive                         | Basic/Primitive/primitive.pd    |
+  | PDL::Slices                            | Basic/Slices/slices.pd          |
+  | PDL::Ufunc                             | Basic/Ufunc/ufunc.pd            |
```
mohawk2, line 175: This section is a maintenance nightmare that might go out of date almost immediately. Can you remove the "not in this repo" ones, and keep it to only ones under `Basic`? Also, the whole thing can probably be replaced with advice to use `git grep PACKAGENAME`.

Basic/Pod/QuickStart.pod
```diff
@@ -41,17 +41,15 @@ data to be easily played with and visualised.
 
 =head2 Help
 
-PDL contains extensive documentation, available both within the
+PDL contains extensive user documentation, available both within the
```
mohawk2, line 44: Probably also worth mentioning MetaCPAN these days! I use `perldl`'s `help (or ?)` function, sometimes `perldoc`, and the web. I've never used `pdldoc`.

mohawk2 (changes requested):
Thank you for this! Once it's updated and merged, could I ask you to comment on the open issues that you've pulled info from? I can then update them to show the points that have been incorporated.

Basic/Pod/DeveloperGuide.pod
HaraldJoerg: I'll remove the "not in this repo ones".  However, I don't think that the information is very volatile. I would actually welcome changes which push the layout into a CPAN mainstream layout and volunteer to adapt the table accordingly.
This table enables me to directly jump to the source file of a package in my editor while `git grep PACKAGENAME` gives false positives.

Basic/Pod/QuickStart.pod
HaraldJoerg: I will add a link to MetaCPAN.  I'll also add a line describing `pdldoc` as the equivalent of `perldoc -f` which in my opinion makes a lot of sense.

mohawk2:
Thank you! Are you happy for me to merge this, or is there more you'd like to change?

mohawk2:
Thank you!

PR #492, opened 2024-08-18
don't execute random perl code in forked child processes

dk:
see the use case in https://github.com/dk/Prima/issues/114

### DIFF
--- a/perldl
+++ b/perldl
@@ -218,14 +218,15 @@ push @enabled, "MultiLines";
 
 my $interrupt_handle = eval 'use Sys::SigAction qw(set_sig_handler); 1';
 my $interrupt_msg = "Ctrl-C detected";
+my $pid = $$;
 if ($interrupt_handle) {
   push @enabled, "Interrupt";
 } else {
-  $SIG{'INT'} = sub { die "$interrupt_msg\n" }; # Ctrl-C handler
+  $SIG{'INT'} = sub { die "$interrupt_msg\n" if $$ == $pid }; # Ctrl-C handler
 }
 sub mksighandle {
   $interrupt_handle ? set_sig_handler(INT => sub {
-    die "$interrupt_msg\n";
+    die "$interrupt_msg\n" if $$ == $pid;
   }) : undef;
 }
 


mohawk2:
Does the change you made in Prima (restoring the actions on signals) make this redundant now?

dk:
No I don't think this is redundant unless you want to have this problem again and again in every subprocess forked from inside perldl. Do you need a formal merge request with a patch or you can copy-paste the fix from my last comment?

dk:
Ah there was already a merge request sorry. I'm not sure why it is stalled then

mohawk2:
My starting point is that PDL (specifically, its REPL, `perldl`) never had a problem. It was not doing anything unreasonable with signals, nor with its stdio. Another module, Prima, was doing something that created a problem, and from what I see on that other issue (https://github.com/dk/Prima/issues/114), seems to have done something to stop causing that problem.

Since you were in fact able to reproduce the problem that Prima was causing for PDL, do the changes you made in Prima now mean that `perldl` doesn't get the errors it did?

dk:
Okay lets get this straight: do you believe that the signal action a process installs, should _not_ exercise a sanity check about its getpid() context?

mohawk2:
I think a process has the right to expect that child processes will not interfere with its stdio. I think that a library that does a `fork` (very especially when that `fork` is not at all fundamental to its operation) has the responsibility to not cause problems through that forking.

dk:
I'm not going to argue about the responsibility as this is philosophical rabbit hole and who should and should not do will lead to endless discussions. As for the technical details, Prima did not mess with perldl stdio, I believe you have this part wrong. So if you believe that a signal handler should not exercise extreme cautions, as it is normal unix prigramming practice in decades, I believe further discussion wouldn't leave anywhere

mohawk2:
By forking and not closing stdin, Prima was unfortunately causing an effect on `perldl`'s stdio. If that were not the case, we wouldn't be having this discussion.

If Prima is still causing a problem in this respect, then the Prima demo will have to be removed, since it causes a terrible user experience.

dk:
There is a direct proof that this problem is caused by improper signal handling. I don't understand where the stdin theory comes from, and threatening to drop Prima support doesn't change that evidence, sorry. But if you are positive that stdin is problematic, you are welcome to submit a demonstration, though without perldl or PDL, so it will be a clean-cut case.

mohawk2:
1) I am not threatening anything. I am simply not going to continue publishing a demo that causes horrible error messages. Do you feel that you have a *right* to keep this demo published that I am "threatening"?

2) What improper signal handling are you saying PDL / `perldl` is doing? Please make clear what makes it "improper". If you wish to claim that it's obvious that Prima is doing `fork`, please highlight where this is documented, and why it is necessary for it to do so.

3) To help me understand what you're saying, please show this "direct proof that this problem is caused by improper signal handling". The "stdin" point is simply because I don't understand why Prima a) needs to do this fork thing at all, but that's not my problem; b) it is my surmise that failing to close `stdin` is what causes the horrible error messages, but I could be wrong. `perldl` will not be changed to accommodate Prima unless and until I see convincing evidence that it is currently incorrect.

HaraldJoerg:
I apologize for jumping onto this discussion...

First let me say that I never saw ugly error messages from the combination of Prima and PDL which I use daily, quite often in the interactive shell.  Therefore, my capability to reproduce the problem are very limited.   What I can reproduce is the following:
```
haj@hajbuntu:~$ perldl
perlDL shell v1.357
 PDL comes with ABSOLUTELY NO WARRANTY. For details, see the file
 'COPYING' in the PDL distribution. This is free software and you
 are welcome to redistribute it under certain conditions, see
 the same file for details.
ReadLines, NiceSlice, MultiLines, Interrupt  enabled
Reading PDL/default.perldlrc...
Found docs database /usr/local/lib/x86_64-linux-gnu/perl/5.38.2/PDL/pdldoc.db
Type 'help' for online help
Type 'demo' for online demos
Loaded PDL v2.099 (supports bad values)

Note: AutoLoader not enabled ('use PDL::AutoLoader' recommended)

pdl> require Prima

Ctrl-C detected
pdl> demo prima
```

So I start `perldl` and say `require Prima` at the first prompt.  This creates a second process (Obviously, I am on one of the platforms where this happens).

Then I hit Ctrl-C which results in the message `Ctrl-C detected` - and at this point, the second process gets defunct.

Nonetheless, the Prima demo starts and works quite fine, including the ability to export a plot (which I _guessed_ would need the second process).  Apparently the demo does not need the second process.

If there is any chance that the death of the second perl process is responsible for the problems, then an easy fix is to call `(void) setsid();` as the first action of the child process ([here](https://github.com/dk/Prima/blob/a826a80685a4aa0595777fd48cf2d46fa85c275d/unix/gtk.c#L320)).  This makes sure that the child process is in its own group and does not receive signals sent to the parent, but will react to signals sent to its own PID.

I verified that this keeps the child process alive when I send Ctrl-C to `perldl` but ... unfortunately I can not spot any different behavior of the demo, which seems to work with and without the patch.

Sorry again for the noise...

dk:
Hello Harald, the problem with inheriting signals in Prima's forked code was solved already some time ago, using explicit sigaction calls; setsid is an elegant option too. I believe though that it is also a good idea for perldl to exercise caution and check getpid() too, just to be more robust; a fairly simple and straighforward proposition, I think. But now as the thread suddenly veers into stdin and who has rights to what and finally threats to remove the demo ... I think I'll just let the PDL people themselves to deal with this, while I'll bow out.

mohawk2:
@dk Thank you.

Do you know if/when you'll ever release a version of Prima that unbreaks working with Strawberry Perl 5.32? I gather from an issue on Prima that it's been fixed there for a while, but for some reason it's not on CPAN.

dk:
1.76 will be there after polishing the new grid geometry manager, in a month or two

HaraldJoerg:
@dk: I am aware of [`restore_signal`](https://github.com/dk/Prima/blob/0980265074ddf72e093a13158cffb470584e9694/unix/gtk.c#L216).  This is good and necessary because the forked task might not be able to process custom signal handlers.  But this routine resets signal handlers to their defaults `SIG_DFL`. This means that a SIGINT will definitely terminate the process, leaving it as a zombie.  So, if any Prima application uses custom signal handlers, for example if I run the application under the Perl debugger, actually sending the signals will kill the forked process.
I still have not found any Prima feature which is not available while the forked process is defunct or even outright killed.  Does this only happen under Wayland?

PR #509, opened 2024-12-14
PDL::IO::Dumper: better handling of multiply referenced ndarrays

### LINKED ISSUE #508: PDL::IO::Dumper frestore doesn't resolve duplicated ndarrays
If one uses fdump to save a data structure that has identical ndarrays (or possibly other objects), then fdump will recognize this and save space in the output file by having the "first" ndarray that was written be referenced by the following ndarrays.  This is great, but it presents a problem when using frestore, I think because it triggers "my variable %s masks earlier declaration in same scope" warnings.  Simple example script attached.
[dumper_collision.pdl.txt](https://github.com/user-attachments/files/18083405/dumper_collision.pdl.txt)

shawnlaffan:
* Keep a track of which ndarrays have already been dumped to avoid duplicating them in the dump string.
 * Handle multiple refs in the dumper structure.  
 * Generalise dumper tests to handle future changes to dump method thresholds.  

Includes some modernisation of code.  More could be done but that's better as a separate PR.  

Fixes #508

### DIFF
--- a/lib/PDL/IO/Dumper.pm
+++ b/lib/PDL/IO/Dumper.pm
@@ -491,7 +491,30 @@ string.  You shouldn't call this unless you know what you're doing.
 sub PDL::IO::Dumper::find_PDLs {
   my($sp, @items) = @_;
 
-  my $out = "";
+  my $out_aref = _find_PDLs_inner($sp, @items);
+
+  #  deduplicate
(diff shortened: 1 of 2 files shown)


mohawk2:
Superb, thanks! Seeing the test code tells me I need to adjust that to use Test::PDL as well.

PR #512, opened 2025-01-02
Fix spelling errors.

sebastic:
The lintian QA reported spelling errors for the Debian package build:

 * compresed -> compressed
 * doesnt    -> does not
 * temproary -> temporary

Note that the "doesn't" typo is caused by single quotes being stripped.


mohawk2:
Thank you!

PR #513, opened 2025-01-02
Fix POD ERRORS.

sebastic:
The lintian QA tool reported a pod-conversion-message issue for the Debian package build:
```
 Around line 184:
     Non-ASCII character seen before =encoding in '“Cygwin'. Assuming UTF-8
```
Using the ASCII double quote characters resolves the issue.


mohawk2:
Thank you also!

PR #520, opened 2025-01-11
Add FREEZE and THAW methods for serialisation

shawnlaffan:
Tests work with Serial, JSON::MaybeXS, CBOR::XS ~and CBOR::Free~.  (Edit: `CBOR::Free` was a false positive - it does not support FREEZE/THAW).

I also had a look at testing with some of the YAML modules but they do not support FREEZE/THAW callbacks.  

Could probably be squashed into a single commit before merging, thus cleaning up debugging cruft.  

Documentation is still needed.  It should just comprise a note that serialisers that conform to `Types::Serialiser` are (should be) supported.  

I'm not sure where to add it, though, as it does not fit under `PDL::IO::Storable`.  Perhaps its contents could be shifted to a new PDL::IO::Serialisation or similar which is then loaded by default with the rest of PDL.  That might be too large a change, though.

### DIFF
--- a/Changes
+++ b/Changes
@@ -16,6 +16,7 @@
 - "usage" section of xform docs now automatically generated (#519)
 - overloaded operators now throw exception if given undef argument (#519) - thanks @cgtinney for report
 - Core::dog now scales linearly with length of top dim, not O(n^2) (#421) - thank @djerius for report
+- PDL::IO::Storable now also supports Sereal, JSON::MaybeXS, CBOR::XS (#510,#520) - thanks @shawnlaffan
 
 2.098 2025-01-03
 - fix Windows build problems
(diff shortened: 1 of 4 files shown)


mohawk2:
I feel like PDL::IO::Storable is a good place - it's just a vehicle to add PDL methods, which is what's needed here? Why do you say it doesn't fit?

Thank you for your great work on this! Could I ask you to add the doc notes you mention?

Don't worry about the commits, I'll do a squash. Could I also ask you though to put the tests in a section in `t/storable.t` instead of a separate file?

shawnlaffan:
The reason for changing things is simply the general vs the specific.  Storable is a specific serialisation method.

mohawk2:
This is great - thank you!

mohawk2:
I've rebased this, and had to make a couple of tweaks (see my commit changes - JSON::PP doesn't work). As soon as the CI blesses it, I'll merge it.

mohawk2:
Thanks again! It looks like your repo has a number of branches you may not need anymore. Time for a spring clean? :-)

Issue #518, opened 2025-01-07
Add benchmarks to check performance


duffee:
Test PDL against [Julia's loop fusion](https://discourse.julialang.org/t/programming-language-benchmark-2/108429)

Comparison with [Programming Language Benchmark v2](https://github.com/attractivechaos/plb2) will require an implementation of **nqueens**, **sudoku** and **bedcov**. I assume we can already do **matmult** out of the box. Rolling this out will go some ways towards covering our collective blushes when it comes to Perl being the slowest on the graph.

We have timing code under **Basic/examples/Benchmark/** already. I think the new examples could live there, unless there's a reason to build more infrastructure that doesn't belong in the core project. Also, follow up on TODOs in documentation for **PDL::MatrixOps** `determinant` v `det` and all of **PDL::Matrix** and try to address the benchmarking comment mentioned in [PDL::Indexing](https://metacpan.org/pod/PDL::Indexing#How-to-figure-out-the-loop-dimensions).

Assign this to me. I've [used Hyperfine before](https://dev.to/duffee/faster-tetranucleotide-k-mer-frequencies-4pnf), but I'll be in touch for the more esoteric bits of C and XS.

## TODO
Write PDL benchmarks for:
- [ ] nqueens
- [ ] matmult
- [ ] sudoku
- [ ] bedcov

and then
- [ ] Identify functions with the longest dwell times
- [ ] Write a guide for how we generate our benchmarks

### Stretch goals
* optimise the current [perl benches in plb2](https://github.com/attractivechaos/plb2/tree/master/src/perl) 
* upload the new PDL benches to [pdl2](https://github.com/attractivechaos/plb2).

mohawk2:
I'm assuming that last one is a Freudian slip? You put "pdl2" ;-)

There's an instant win for the one Perl one I looked at, `nqueens`. Near the start, it first initialises some Perl arrays, *then* loops through initialising them (even though they already have the right values). Just deleting that loop can only save time.

And yes, the whole point of making these benchmarks in PDL (that perform well) would be to PR them into that repo. I'm pretty sure with auto-pthreading, the PDL version would (for algorithms that *can* be parallelised, which my first glance at `nqueens` did not show) actually be faster than C. That was my finding with https://github.com/Fourmilab/floating_point_benchmarks/pull/1

See #451 for performance-measuring tools and techniques.

Another target: https://github.com/FPBench/FPBench

duffee:
[...]

PR #526, opened 2025-01-21
Adding the matmul benchmark from plb2

duffee:
Getting a start on #518. Adding **matmul** from the [Programming Language Benchmark](https://github.com/attractivechaos/plb2), I've created a **tasks** directory under **examples/Benchmark** for the scripts and included a guide to hold notes for hopeful benchmarkers.

If there's a better way of creating the matrices, let me know.

You can run a comparison with the [Perl script](https://github.com/attractivechaos/plb2/blob/master/src/perl/matmul.pl) using [hyperfine](https://github.com/sharkdp/hyperfine)
```
hyperfine --warmup 3 'path/to/matmul.pl 300' 'path/to/matrix_multiplication.pl 300'
```
I recommend low values at first because at N=1500, hyperfine takes over 10 minutes to measure the Perl script.


mohawk2 (changes requested):
Please could you add the note you included in the comment here (about `hyperfine`) to the guide you made? And rename that guide to `README.md`, since that's effectively what it is?

mohawk2:
You'll note this is failing one of the release tests, which you can repro locally with `prove -l xt`. The solution is to run `rm MANIFEST; make manifest`

duffee:
Adding notes to the new README.md:

I need a suggestion on how other devs can test two different versions of PDL on the same benchmark on the same command line.  Different environment variables are a possibility, as are different paths.  How do you work with 2 PDLs in the same git repo?  Hyperfine takes the commands to run in single quotes, like so

`hyperfine 'DEV_ENV=1 PERL5_LIB=over/here ./matrix_multiplication.pl' '...'`

Thoughts?

mohawk2:
> How do you work with 2 PDLs in the same git repo?

It's not really a "thing". Why would it be?

By the way, your most recent commit is a merge instead of a rebase. Please rebase and force-push, like the other times. Please make this part of your normal practice.

mohawk2:
> You'll note this is failing one of the release tests, which you can repro locally with `prove -l xt`. The solution is to run `rm MANIFEST; make manifest`

This is still not fixed.

duffee:
For 2 PDLs, I'm thinking of the case where you want to know if your code change has affected the benchmark.

I've added this to my git config
```
[core]
       pull.rebase = true   
```

MANIFEST is fixed locally and will go up in the next commit. If you've no bright ideas, I'll push the README I've got so far.

mohawk2:
> For 2 PDLs, I'm thinking of the case where you want to know if your code change has affected the benchmark.

In the normal way, I'd run the benchmark, make the change, run it again. But one can achieve similar with `git checkout HEAD^` or any other ref. This is another use-case for `git bisect`, of course, to find where things changed. If Hyperfine is really taking 10 mins to speed-check a simple `matmult`, is it really worth using?

> MANIFEST is fixed locally and will go up in the next commit. If you've no bright ideas, I'll push the README I've got so far.

My approach would be (and is) to push changes as I have them. Small, beneficial changes should be the "unit" of a commit.

duffee:
Requested changes made to MANIFEST and README.md

PR submitted to [attractivechaos/plb2](https://github.com/attractivechaos/plb2/pull/82)

mohawk2:
This PR is now in much better shape, thank you. Your PR on plb2 looks good, too.

Are you planning to add more to this, or shall I merge it?

duffee:
You can merge this. I'll come back to the other benchmarks after I've got my modules passing cpants again.

PR #529, opened 2025-01-24
Core.pm: Fix POD heading for update_data_from

HaraldJoerg:


HaraldJoerg:
I looked for `pdldoc update_data_from` in 2.099 and it wasn't there... but the fix is trivial

mohawk2:
Thank you!

PR #532, opened 2025-01-31
Allow an ndarray as Center in rvals

wlmb:


mohawk2 (changes requested):
Thanks, this looks really good. Please can you make it a `die` or `confess` if too many elements are supplied? I'm keen to not allow errors to pass through and cause problems later.

wlmb:
> Thanks, this looks really good. Please can you make it a `die` or `confess` if too many elements are supplied? I'm keen to not allow errors to pass through and cause problems later.

Done. Would it be better to 'use Carp' and 'croak' instead of 'die'?

mohawk2:
> > Thanks, this looks really good. Please can you make it a `die` or `confess` if too many elements are supplied? I'm keen to not allow errors to pass through and cause problems later.
> 
> Done. Would it be better to 'use Carp' and 'croak' instead of 'die'?

I've just taken a look, and because there's a `use PDL::Core ''`, `barf` is already imported (and is used elsewhere in the file). Please use that (that should have been my first ask, lesson for both of us there :-).

wlmb:
I added a check on the dimensions of the center (it should be 1D) and changed warn to die to barf. I used topdl to unify the cases of a reference or of an ndarray and then converted to a perl array in order not to change the rest of the code.

mohawk2:
Looks really good! One more nitpick: `PDL::Core::topdl($opt{CENTRE})` should be `PDL->topdl($opt{CENTRE})`.

mohawk2:
By the way, you wouldn't happen to know about doing ANOVA with generalised linear modelling (GLM)? I've got *one* case that doesn't work in https://rt.cpan.org/Ticket/Display.html?id=97925, having fixed the other? The problem I'm having is the remaining case gives a singular matrix when effect-coded up, and I'm grinding through the textbook to see if there are known techniques to avoid such singularity. If not, no problem!

wlmb:
> Looks really good! One more nitpick: `PDL::Core::topdl($opt{CENTRE})` should be `PDL->topdl($opt{CENTRE})`.

Done.

wlmb:
> By the way, you wouldn't happen to know about doing ANOVA with generalised linear modelling (GLM)? I've got _one_ case that doesn't work in https://rt.cpan.org/Ticket/Display.html?id=97925, having fixed the other? The problem I'm having is the remaining case gives a singular matrix when effect-coded up, and I'm grinding through the textbook to see if there are known techniques to avoid such singularity. If not, no problem!

Sorry. I'm ignorant about ANOVA. Just read a little bit and seems interesting. A wild guess is that SVD (singular value decomposition) might be useful : it allows solving, in a sense, some singular problems. Regards, Luis

PR #539, opened 2025-04-13
pptemplate improvements

HaraldJoerg:
As discussed on IRC:

- Minor bugfix: Requires 'PDL' instead of 'PDL::Basic'
- The script now creates MANIFEST, MANIFEST.SKIP and .gitigignore  files for bookkeeping
- The script writes one line what it does
- POD is adapted to the new "deep" mode results

The second commit recommends pptemplate in the documentation of PDL::PP.

Further ideas:

- The error messages of pptemplate are rather dry.  Using [autodie](https://perldoc.perl.org/autodie) would bring an improvement, but I know that many people are uncomfortable _not_ checking the return code of open etc.
- I did a manual test which could be automated as an xt test:
  1 pptemplate PDL:Some:Module
   2 cd PDL-Some-Module  
   3 git init
   4 git status > /tmp/before
   5 perl Makefile.PL
   6 make
   7 make distcheck # should pass, i.e. not mention MANIFEST
   8 make dist
   9 git status > /tmp/after
   10 git diff /tmp/before /tmp/after # must be empty

### DIFF
--- a/script/pptemplate
+++ b/script/pptemplate
@@ -1,4 +1,4 @@
-#!perl -w
+#!/usr/bin/env perl
 
 use strict;
 use warnings;
@@ -90,13 +90,13 @@ WriteMakefile(
   MIN_PERL_VERSION => '5.016',
   LICENSE=> 'perl',
   PREREQ_PM => {
-    'PDL::Basic'  => '2.096', # deep mode
+    'PDL'  => '2.096', # deep mode
   },
   CONFIGURE_REQUIRES => {
-    'PDL::Basic'  => '2.096',
+    'PDL'  => '2.096',
   },
   BUILD_REQUIRES => {
-    'PDL::Basic'  => '2.096',
+    'PDL'  => '2.096',
   },
   TEST_REQUIRES => {
     'Test::More' => '0.88', # done_testing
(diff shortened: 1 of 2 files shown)


script/pptemplate
```diff
@@ -117,6 +117,80 @@ sub postamble { ::pdlpp_postamble(\@pd_srcs) }
...
+sub pdManifest_skip {
+  my ($pdname,$dir) = @_;
+  my $pd_dir = dirname $pdname;
+  my $pm_name = $pdname =~ s/\.pd$/.pm/r;
+  return <<"SKIP_PROJECT" . <<'SKIP_GENERAL';
+# This project's temporary files
+$pm_name
+$pd_dir/.*-pp-.*
+$pd_dir/.*\\.bs\$
+$pd_dir/.*\\.c\$
```
mohawk2, line 141: This is too broad, since the developer might add their own C files, for example https://github.com/PDLPorters/pdl/blob/master/lib/PDL/Primitive-xoshiro256plus.c - there's a reason I asked you to use the pattern `^lib/PDL/Compression(\.(pm|xs|c)$|-pp-)` :-)

script/pptemplate
```diff
@@ -1,4 +1,4 @@
-#!perl -w
+#!/usr/bin/env perl
```
mohawk2, line 1: This change is incorrect. EUMM updates it to be exactly the same Perl as the one installing it. Please leave it as is.

mohawk2 (changes requested):
Thank you, this is great! A couple of change are needed.

script/pptemplate
HaraldJoerg: Ah, I seem to have missed that recommended pattern.  Was that on IRC? I'll fix that.

script/pptemplate
HaraldJoerg: Ok, I'll revert that!

PR #540, opened 2025-04-17
Cleanup PDL::IO::FITS tests, fix incorrect tests; fix regression in writing 0D ndarrays

djerius:
`t/fits.t` reorganization and fixes:

- many tests shared variables, requiring explicit re-initialization
- some regression tests for legacy FITS header writing were incorrectly excluded if `Astro::FITS::HEADER` was installed
- the test for tilde expansion in file names was never run.
- temporary files were explicitly unlinked instead of using the automatic facilities of `File::Temp`

Tests are now isolated into subtests, with no shared variables. The other issues were dealt with, but the level of header checking with and without `Astro::FITS::Header` is still quite asymmetric.

bug fixes:

- null hdus were always written using the legacy Perl code due to an incorrect check if `Astro::FITS::Header` was available.

regression fixes:

- passing 0D ndarrays to `wfits` caused an internal crash, as the code
  expected they be at least 1D. This used to work.

### DIFF
--- a/t/fits.t
+++ b/t/fits.t
@@ -77,8 +77,8 @@ subtest 'Astro::FITS::Header' => sub {
             is( $$table2{hdr}{TTYPE2}, "COLB", "column #2 is COLB" );
             is( $$table2{hdr}{TFORM2}, "1D",   "  stored as 1D" );
 
-            is_pdl $x, $$table2{COLA}, "COLA";
-            is_pdl $y, $$table2{COLB}, "COLB";
+            is_pdl $$table2{COLA}, $x, "COLA";
+            is_pdl $$table2{COLB}, $y, "COLB";
 
         };
 
@@ -109,8 +109,8 @@ subtest 'Astro::FITS::Header' => sub {
             is( $$table2{hdr}{TTYPE2}, "BAR", "column #2 is BAR" );
             is( $$table2{hdr}{TFORM2}, "1J",  "  stored as 1J" );
 
-            is_pdl $x, $$table2{BAR}, "BAR";
-            is_pdl $y, $$table2{FOO}, "FOO";
+            is_pdl $$table2{BAR}, $x, "BAR";
+            is_pdl $$table2{FOO}, $y, "FOO";
         };
 
         # try out more "exotic" data types
--- a/t/fits.t
+++ b/t/fits.t
@@ -418,4 +418,41 @@ subtest 'multi-line HISTORY' => sub {
 
 };
 
+###############################
+subtest 'write null hdu with and without Astro::FITS::Header' => sub {
+
+    subtest 'with' => sub {
+      SKIP: {
+            skip 'Astro::FITS::Header not available'
+              unless $PDL::Astro_FITS_Header;
+            my ( $fh, $file ) = tfile;
+            my $x = pdl(3);
+            lives_ok { wfits [pdl([3])], $file } 'create file';
+
+            my $contents = do {
+                local $/;
+                open my $fh, '<', $file
+                  or die("unable to open $file");
+                <$fh>;
+            };
+            unlike( $contents, qr/legacy code/, "didn't use legacy code" );
+        }
+    };
+
+    subtest 'without' => sub {
+        local $PDL::Astro_FITS_Header = 0;
+        my ( $fh, $file ) = tfile;
+        my $x = pdl(3);
+        lives_ok { wfits [pdl( [3] )], $file } 'create file';
+        my $contents = do {
+            local $/;
+            open my $fh, '<', $file
+              or die("unable to open $file");
+            <$fh>;
+        };
+        like( $contents, qr/legacy code/, "used legacy code" );
+    };
+
+};
+
 done_testing();
--- a/t/fits.t
+++ b/t/fits.t
@@ -455,4 +455,27 @@ subtest 'write null hdu with and without Astro::FITS::Header' => sub {
 
 };
 
+###############################
+subtest 'accept scalar (0D) ndarrays' => sub {
+
+    subtest 'bintable' => sub {
+        my ( $fh, $file ) = tfile;
+        lives_ok { wfits { x => pdl(3) }, $file } 'write column';
+        my $got = rfits($file);
+        is_pdl( $got->{X}, pdl( [3] ), "got data" );
+
+    };
+
+    subtest 'image' => sub {
+        my ( $fh, $file ) = tfile( UNLINK => 0 );
+        diag $file;
+        lives_ok { wfits pdl(3), $file } 'write image';
+        my $got;
+        lives_ok { $got = rfits($file) } 'read image';
+        is_pdl( $got, pdl( [3] ), "got data" );
+
+    };
+
+};
+
 done_testing();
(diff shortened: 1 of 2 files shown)


t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+        my $hist = $m51r->hdr->{'HISTORY'};
+        $hist = join '', map "$_\n", @$hist if ref $hist eq 'ARRAY';
+        $hist =~ s/ +$//gm;
+        is( $hist, $hstr, 'correct with pre-existing header' );
+    };
+
+};
+
+###############################
+subtest 'write null hdu with and without Astro::FITS::Header' => sub {
```
mohawk2, line 422: The existence of `t/fits-noafh.t` renders this whole section moot. In particular, merely checking that `wfits` didn't throw an exception seems to miss the point a bit. Please remove this change, or make it unconditional on AFH being there, and do an actual check of the read-back data.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+            is( $$table2{tbl}, "binary", "and appears to be a binary TABLE" );
+
+            ok( exists $$table2{COLA} && exists $$table2{COLB},
+                "columns COLA and COLB exist" );
+            is( $$table2{hdr}{TTYPE1}, "COLA", "column #1 is COLA" );
+            is( $$table2{hdr}{TFORM1}, "1J",   "  stored as 1J" );
+            is( $$table2{hdr}{TTYPE2}, "COLB", "column #2 is COLB" );
+            is( $$table2{hdr}{TFORM2}, "1D",   "  stored as 1D" );
+
+            is_pdl $x, $$table2{COLA}, "COLA";
```
mohawk2, line 80: These two have the "got" and "expected" values the wrong way round. Please fix this.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+                    BAR => $x,
+                    FOO => $y,
+                    hdr => { TTYPE1 => 'FOO', TTYPE2 => 'BAR' }
+                };
+                my ( $fh, $file ) = tfile;
+                wfits $table, $file;
+                rfits $file;
+            };
+
+            ok(
```
mohawk2, line 101: Please do here the multiple checks you did above. A single "not OK" would be a rather useless test diagnostic.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+        subtest 'exotic data types' => sub {
+            my $x = byte( 12, 45, 23, 0 );
+            my $y = short( -99, 100, 0, 32767 );
+            my $c = ushort( 99, 32768, 65535, 0 );
+            my $d = [ "A string", "b", "", "The last string" ];
+            my $e = float( -999.0, 0, 0, 12.3 );
+            ##my $f = float(1,0,-1,2) + i * float( 0,1,2,-1 );
+
+            my $table2 = do {
+                my $table = {
```
mohawk2, line 127: This intermediate variable that's used once is not adding value. Please just call `wfits` with that value.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+                    DCOL => $d,
+                    ECOL => $e,
+                    ##	  FCOL => $f,
+                };
+                my ( $fh, $file ) = tfile;
+                wfits $table, $file;
+                rfits $file;
+            };
+
+            ok(
```
mohawk2, line 140: Multiple checks again please.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
+        my @target_bitpix = ( 8, 16, 32, -32, -64 );
+        my $bp_i          = 0;
+        for my $cref ( \( &byte, &short, &long, &float, &double ) ) {
+            for my $x ( $a1, $a2 ) {
+                $p = &$cref($x);
+                my ( $fh, $file ) = tfile;
+                $p->wfits($file);
+                $q = PDL->rfits($file);
+                my $flag = 1;
+                if ( ${ $p->get_dataref } ne ${ $q->get_dataref } ) {
```
mohawk2, line 219: A simple `is` here would suffice. Please use that.

t/fits.t
```diff
@@ -2,340 +2,480 @@ use strict;
...
-if(-w dirname($tildefile)) {
-	lives_ok { sequence(3,5,2)->wfits($tildefile) } "wfits tilde expansion didn't fail";
-	lives_ok { rfits($tildefile) } "rfits tilde expansion didn't fail";
-	$tildefile =~ s/^(~)/glob($1)/e; #use the same trick as in FITS.pm to resolve this filename.
-	unlink($tildefile) or warn "Could not delete $tildefile: $!\n"; #clean up.
-}
+  SKIP: {
+        my $tilde_dir = <~>;
+
+        skip '~ directory is not writeable'
```
mohawk2, line 339: This is an incorrect call to `skip`.

mohawk2 (changes requested):
Please make the changes noted. It's a bit surprising that you submitted this without (from the CI results) even running it locally.

djerius:
> Please make the changes noted. It's a bit surprising that you submitted this without (from the CI results) even running it locally.

Harsh, dude. No need for that.

t/fits.t
djerius: It does do an actual check of the readback data.  the test for exception is secondary to the check which appears on the next lines.

t/fits.t
djerius: ok.

t/fits.t
djerius: yup.

t/fits.t
mohawk2: OK. Then please get rid of this whole bit, since exceptions make tests fail anyway.

t/fits.t
djerius: ok

t/fits.t
djerius: ok

t/fits.t
djerius: Looks like Test2's skip doesn't require the count, which is why it passed on my end. fixed.

djerius:
> > regression fixes:
> > 
> > * passing 0D ndarrays to `wfits` caused an internal crash, as the code expected they be at least 1D. This used to work.
> 
> Can you confirm in which version that used to work? It will need noting in `Changes` (which I'll update).

I know with certainty it worked back in 2.018. I'll try to track down the version where it changed.

djerius:
> > > regression fixes:
> > > 
> > > * passing 0D ndarrays to `wfits` caused an internal crash, as the code expected they be at least 1D. This used to work.
> > 
> > 
> > Can you confirm in which version that used to work? It will need noting in `Changes` (which I'll update).
> 
> I know with certainty it worked back in 2.018. I'll try to track down the version where it changed.

git bisect says 899a6159a60478dd0d197a1623981237efade1d6.

git run script:

```
git reset --hard
perl Makefile.PL
make -j80 && perl -Mblib -MPDL -MPDL::IO::FITS=wfits -E 'wfits { a => pdl(3)}, q{/tmp/foo}'
```

Issue #541, opened 2025-05-05
Build is not reproducible


daym:
The build of pdl is not reproducible.

The pdldoc.db contents keep changing on builds.

That's because File::find does not order its result, among other things. That means results of File::find are not in a stable order.

daym:
Thanks!

By now, I got reproducibility up from 4% to 64.7%.

I would like some help with the rest. I didn't write the doc generator so I wouldn't know where to look for generators of randomness in its source code.

mohawk2:
If you put on here things that differ across builds, I'm pleased to assist with pointers where they come from.

daym:
Differences:

[Q.txt](https://github.com/user-attachments/files/20047927/Q.txt)

That is after I did `od -tx1 -c` on both variants of `pdldoc.db`--which was necessary since it's a binary file.

mohawk2:
I'm expecting you to identify the differences, not to have to `diff` files myself. Look for any examples of `keys`, `values` and `each`, they are sources of randomness.

daym:
I don't know what that means. I actually had a file with differences attached a minute ago and removed it again since it's too user-unfriendly. OK, I'll add the things that differ across builds now after all:

Those things differ: [Q.txt](https://github.com/user-attachments/files/20047990/Q.txt)

I don't know what it's file format is, so I cannot decode it.

mohawk2:
The colossal file you've attached just tells us that all the lines in the two big files differ. You may think that's useful info, but I don't.

The file format is a bespoke one that's just null-separated in tuples. Search for the word `join`, there are exactly 3 occurrences in that file. The order of the arg to that might provide the rest of the answer.

daym:
Thanks!

With the following additional change (also see PR), it builds reproducibly.

```
-      my $txt = join(chr(0),$name,$module,%$val);
+      my $txt = join(chr(0),$name,$module,map { "$_" . chr(0) . "$val->{$_}" } sort keys %$val);
```

Not sure whether it's still the same format as before, though.

mohawk2:
Please think about whether there is a difference between `%$val` and `sort keys %$val`.

While you're doing that, please try installing your modified version, then using the `help` function in `perldl`.

daym:
I've tried the following test and they seem equal to me:

```
my $name = "example_name";
my $module = "example_module";
my $val = {
    key1 => "value1",
    key2 => "value2",
    key3 => "value3",
};

[...]

Issue #349, opened 2021-11-14
Support also doing processing on GPUs


mohawk2:
This was discussed on pdl-devel back in Dec 2015 (see https://sourceforge.net/p/pdl/mailman/pdl-devel/?viewmonth=201512). There is also some discussion in Jun this year on PerlMonks (https://www.perlmonks.org/?node_id=11134476).

I recently revisited this; my thinking so far is that the best platform for mixed CPU/GPU processing seems to be OpenCL. However, OpenCL C doesn't support native-complex processing, but it would be possible to use OpenCL C++ with a class that overloads the arithmetic operators for source compatibility with existing `Code`; an easy first-cut alternative would be to only support GPU use on real types (which may also be further restricted to not supporting `double` for <OpenCL 2.0 environments - GPUs have long been aimed at speed rather than extreme precision, because of graphics' needs).

Defining some terms: "kernel" is a term in GPGPU-land for the code that actually runs in the GPU, typically being run once per "work item", which is basically identical to the "inner loop" inside a `threadloop %{ /* blah */ %}` (and/or `loop`). Current PDL operations have code generated by PP that doesn't distinguish between the "kernel" and the setup code (which runs on the "host"), since the output is just a bunch of C to put inside the function. When that distinction starts to matter, it will need some work.

Things that would need doing to allow this:
- split `Code` (etc) into kernel and setup code; another benefit of doing so would be to aid lazy-building ndarrays since PDL could be made to set up a set of very small ndarrays, call the kernel on that, then do something with the results, then throwing them away - an example might be to `sumover` a very long `sequence` using basically no memory
- figure a way to allow calling at least one other different threadloop initiator (as called in PP-generated C code); for back-compat with current CPU-only it's not desirable to make PDL have a hard requirement of OpenCL to operate at all
- implement dividing work into a task queue with 4x as many items as the number of processing-units as alluded to here (https://sourceforge.net/p/pdl/mailman/pdl-devel/?viewmonth=202109) (which will also help with parallelising non-deterministic tasks)
- either figuring how to identify code that calls into GPU-unavailable C functions (e.g. LAPACK) and having that not use GPU, or figure the best way to manage opt-in / opt-out of GPU use - an easy way might be to simply try OpenCL-compiling the generated code, which undoubtedly will fail if any unsuitable function calls (or headers) are found

Resources:
- https://github.com/CastXML/CastXMLSuperbuild which illustrates how to use CMake to build an external project (here, LLVM might be needed)
[...]

PR #542, opened 2025-05-05
Make doc generator reproducible.

daym:
Fixes <https://github.com/PDLPorters/pdl/issues/541>.

### DIFF
--- a/lib/PDL/Doc.pm
+++ b/lib/PDL/Doc.pm
@@ -495,15 +495,17 @@ sub savedb {
   my $hash = $this->ensuredb;
   open my $fh, '>', $this->{Outfile} or barf "can't write to symdb $this->{Outfile}: $!";
   binmode $fh;
-  while (my ($name,$mods_hash) = each %$hash) {
+  for my $name (sort keys %$hash) {
+    my $mods_hash = $hash->{$name};
     next if 0 == scalar(%$mods_hash);
-    while (my ($module,$val) = each %$mods_hash) {
+    for my $module (sort keys %$mods_hash) {
+      my $val = $mods_hash->{$module};
       my $fi = $val->{File};
       $val->{File} = abs2rel($fi, dirname($this->{Outfile}))
         #store paths to *.pm files relative to pdldoc.db
         if file_name_is_absolute($fi) && -f $fi;
       delete $val->{Dbfile}; # no need to store Dbfile
-      my $txt = join(chr(0),$name,$module,%$val);
+      my $txt = join(chr(0),$name,$module,map +($_=>$val->{$_}), sort keys %$val);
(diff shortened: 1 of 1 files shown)


lib/PDL/Doc.pm
```diff
@@ -495,9 +495,12 @@ sub savedb {
   my $hash = $this->ensuredb;
   open my $fh, '>', $this->{Outfile} or barf "can't write to symdb $this->{Outfile}: $!";
   binmode $fh;
-  while (my ($name,$mods_hash) = each %$hash) {
+
```
mohawk2, line 498: Please don't introduce blank lines.

mohawk2 (changes requested):

lib/PDL/Doc.pm
```diff
@@ -679,7 +681,9 @@ sub scan {
   my $hash = $this->{SYMS} ||= {};
   my $n = 0;
   $_->{File} = $file2, $n++ for values %{ $parser->{SYMHASH} };
-  while (my ($key,$val) = each %{ $parser->{SYMHASH} }) {
+
```
mohawk2, line 684: This is still a blank line. I did ask you not to introduce those.

mohawk2 (changes requested):
Obviously please also add the change discussed on #541.

mohawk2:
Thanks! If the CI likes this version, it's getting merged.

May I ask what this is for, by the way?

mohawk2 (approved):

daym:
I'm packaging this version of perl-pdb for GNU Guix. That distribution uses content-addressable storage (like Fedora Core, flatpak etc). Thus having a package whose output keeps changing for no reason would eventually use up all the storage space of the build machine. It's also not good for reproducible science to have mutating tools.

We're using pdl 2.019 in research into and modeling of rare earth magnets and want to see whether this new version of pdl improves parallelism.

mohawk2:
I also doubt that!

As of 2.059, if you have an ndarray of over a configurable size (default 1e6) of elements, then the pthread facility will automatically be used for all calculations. See https://metacpan.org/pod/PDL::ParallelCPU for details. For now, that is still CPU-only, but I am working on adding GPU capability. See #349 for more.

zmughal:
Great work on this! I have been thinking about adopting Guix for certain purposes myself.

I wanted to add that it might be useful to add a comment to that code indicating the purpose of sorting is to output in a canonical order for reproducible builds (since there isn't a test and any future person touching that code could probably figure out why it is there after thinking about it or looking at the log, but this makes it explicit).

mohawk2:
Anyone confused about why the code is as it is would use `git blame`, which would bring up the commit message about reproducibility. Anyone who saw `sort keys` and that but was still confused probably shouldn't be touching the code anyway.

PR #553, opened 2026-04-22
Minor corrections to FAQ

duffee:
Updated perl version to 5.16 in Q2.6, replaced `pp_end` with `pp_done` in Q7.2 and removed whitespace.

### DIFF
--- a/lib/PDL/FAQ.pod
+++ b/lib/PDL/FAQ.pod
@@ -38,7 +38,7 @@ this document but currently aren't.
 
 Similarly, if you think parts of this document are
 unclear, please tell the FAQ maintainer about it. Where
-a specific answer is taken in full from someones posting
+a specific answer is taken in full from someone's posting
 the authorship should be indicated, let the FAQ maintainer
 know if it isn't. For more general information explicit
 acknowledgment is not made in the text, but rather there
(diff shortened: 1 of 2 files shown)


lib/PDL/FAQ.pod
```diff
@@ -1013,7 +1013,7 @@ For another example check the script
 F<t/subclass.t> in the PDL distribution.
 
 
-=head2 Q: 6.15   What on earth is this dataflow stuff ?  
+=head2 Q: 6.15   What on earth is this dataflow stuff?
 
 
 Dataflow is an experimental project that you don't need to concern
```
mohawk2, line 1019: A further FAQ change that updated this would be very welcome! It's now far from experimental.

lib/PDL/FAQ.pod
```diff
@@ -1438,7 +1437,7 @@ C<quickstats.pd> )
                      $avg() = tmp/$SIZE(n);
                   $max() = curmax;
                   $min() = curmin;
-                    '
+                 '
```
mohawk2, line 1440: The examples should probably all use here-docs instead of `'`.

mohawk2:
Thank you for this!

The CI failure was because your branch was 295 commits behind current `master`, and the FreeBSD config has changed since then because the old image went EOL.

PR #555, opened 2026-04-27
Convert tabs to spaces in PP.pod code blocks

duffee:
Tabs in POD were messing up the layout of the code blocks in **PP.pod**. I converted tabs to 4 spaces leaving all other formatting the same. I did see examples of 1-space indents, but left them for now as they could have been intentional.

Will make changes on request.

### DIFF
--- a/lib/PDL/PP.pod
+++ b/lib/PDL/PP.pod
@@ -58,7 +58,7 @@ PDL|https://arxiv.org/abs/1702.07753>.
 
 Why do we need PP? Several reasons: firstly, we want to be able to
 generate subroutine code for each of the PDL datatypes (PDL_Byte,
-PDL_Short, etc).  AUTOMATICALLY.  Secondly, when referring to slices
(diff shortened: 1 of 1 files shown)


lib/PDL/PP.pod
```diff
@@ -58,7 +58,7 @@ PDL|https://arxiv.org/abs/1702.07753>.
 
 Why do we need PP? Several reasons: firstly, we want to be able to
 generate subroutine code for each of the PDL datatypes (PDL_Byte,
-PDL_Short, etc).  AUTOMATICALLY.  Secondly, when referring to slices
+PDL_Short, etc.) AUTOMATICALLY.  Secondly, when referring to slices
```
mohawk2, line 61: Please go with a convention of not having `.` at the end of Latin abbreviations like "etc". For "eg", "ie", I rarely but not never put the two `.` in.

mohawk2 (approved):
Noted a nit for going forward, but this is great, and a definite further improvement. Thank you!

PR #556, opened 2026-04-30
POD cleanup

duffee:
Mostly adding whitespace, but also a TODO to find a missing script from 1998 mentioned in **Tips.pod** and confusion on the first line of **Type.pod**. I've been reading `PDL::Types` and can't figure out what the author was trying to say.

### DIFF
--- a/lib/PDL/QuickStart.pod
+++ b/lib/PDL/QuickStart.pod
@@ -61,7 +61,7 @@ which can hold numbers or strings, lists or arrays of scalars, e.g. C<@x>,
 and associative arrays/hashes of scalars, e.g. C<%x>.
 
 Perl v5 introduces to Perl data structures and objects. A simple
-scalar variable C<$x> now be a user-defined data type or full blown
+scalar variable C<$x> can now be a user-defined data type or full blown
 object (it actually holds a reference (a smart "pointer") to this
 but that is not relevant for ordinary use of perlDL)
 
@@ -101,8 +101,7 @@ PerlDL is loaded into your Perl script using this command:
 
  use PDL;  # in Perl scripts: use the standard perlDL modules
 
-There are also a lot of extension modules, e.g. 
-L<PDL::Graphics::TriD>. 
+There are also a lot of extension modules, eg L<PDL::Graphics::TriD>.
 Most of these (but not all as sometimes it is not appropriate) follow
 a standard convention. If you say:
 
@@ -138,7 +137,7 @@ Here are some ways of creating a PDL variable:
  $c = pdl q[1 - 2];            # 2-element ndarray containing 1 and -2
  $y = pdl 42                   # 0-dimensional scalar
  $c = pdl $x;                  # Make a new copy
- 
+
  $d = byte [1..10];            # See "Type conversion"
  $e = zeroes(3,2,4);           # 3x2x4 zero-filled array
  
@@ -156,13 +155,13 @@ to PDL variables on-the-fly.
 
 =head2 Arithmetic (and boolean expressions)
 
- $x = $y + 2; $x++; $x = $y / $c; # Etc.
+ $x = $y + 2; $x++; $x = $y / $c;     # Etc
 
- $c=sqrt($x); $d = log10($y+100); # Etc
+ $c = sqrt($x); $d = log10($y + 100); # Etc
 
- $e = $x>42; # Vector conditional
+ $e = $x > 42; # Vector conditional
 
- $e = 42*($x>42) + $x*($x<=42); # Cap top
+ $e = 42*($x > 42) + $x*($x <= 42); # Cap top
 
  $y = $x->log10 unless any ($x <= 0); # avoid floating point error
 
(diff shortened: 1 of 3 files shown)


lib/PDL/Tips.pod
```diff
@@ -21,7 +21,7 @@ HTML versions of the pages should also be present, in the
 F<HtmlDocs/PDL> directory of the PDL distribution. To find this 
 directory, try the following
 
- pdl> foreach ( map{"$_/PDL/HtmlDocs"}@INC ) { p "$_\n" if -d $_ }  
+ pdl> foreach ( map {"$_/PDL/HtmlDocs"} @INC ) { p "$_\n" if -d $_ }
```
mohawk2:
```suggestion
 pdl> foreach ( map "$_/PDL/HtmlDocs", @INC ) { p "$_\n" if -d $_ }
```

But in any case, this whole section is moot because we now don't generate HTML docs, so please remove :-)

lib/PDL/QuickStart.pod
```diff
@@ -290,7 +290,7 @@ Simple sectioning uses a syntax extension to Perl,
 L<PDL::NiceSlice>, that allows you to specify subranges
 via a null-method modifier to a PDL:
 
-  $g = $f->($x1:$x2,$y1:$y2,($z1)); # Take subsection
+  $g = $f->($x1:$x2, $y1:$y2, ($z1)); # Take subsection
```
mohawk2: There's a highly informal convention in PDL code for maths-y functions and PDL-specific stuff like slices (as this is) and so on to not have spaces between args. For normal function calls, they should be there. Please don't change away from that.

lib/PDL/QuickStart.pod
```diff
@@ -156,13 +156,13 @@ to PDL variables on-the-fly.
...
+ $x = $y + 2; $x++; $x = $y / $c;     # Etc
 
- $c=sqrt($x); $d = log10($y+100); # Etc
+ $c = sqrt($x); $d = log10($y + 100); # Etc
 
- $e = $x>42; # Vector conditional
+ $e = $x > 42; # Vector conditional
 
- $e = 42*($x>42) + $x*($x<=42); # Cap top
+ $e = 42*($x > 42) + $x*($x <= 42); # Cap top
```
mohawk2, line 164: This whitespace change (unlike the comma-related one below) is nice, thank you.

lib/PDL/QuickStart.pod
```diff
@@ -301,18 +301,18 @@ You can put PDL slices on either side of the element-wise assignment
 operator C<.=>, like so:
 
   # Set part of $bigimage to values from $smallimage
-  $bigimage->($xa:$xb,$ya:$yb) .= $smallimage;
+  $bigimage->($xa:$xb, $ya:$yb) .= $smallimage;
```
mohawk2: See above - not this, please.

lib/PDL/Tips.pod
```diff
@@ -66,13 +66,13 @@ to
...
-forgotten. It is like copy except it changes the original ndarray
-and returns it).
+forgotten. It is like C<copy> except it changes the original ndarray
+and returns it.
 
 Of course, the probably best way to do the above is
 
 	$inds = xvals ($x->long);
-	while(1) {
+	while (1) {
```
mohawk2, line 73: Definitely this is great!

lib/PDL/Type.pod
```diff
@@ -12,7 +12,7 @@ PDL::Type - Objects encapsulating datatypes for PDL transformations
 
 =head1 DESCRIPTION
 
-This module declares one class - of this class The L<type|PDL::Core/type>
+This module declares one class - of this class. The L<type|PDL::Core/type>
```
mohawk2:
```suggestion
This class encapsulates type information for PDL ndarrays. The L<type|PDL::Core/type>
```

mohawk2 (changes requested):
Thank you. Requesting some tweaks.

lib/PDL/Tips.pod
```diff
@@ -84,6 +84,8 @@ which doesn't save all the temporary instances of $x in memory.
 See C<mandel.pl> in the Demos subdirectory of the PerlDL distribution
 for an example.
 
+=for comment TODO: find mandel.pl from 1998 (not in PDL/Demos/ anymore)
```
mohawk2: It might have ended up in the `demo 3dgal`, which has several Mandelbrot demos.

duffee: It is, but for the life of me, I can't find a single reference to it anywhere in the repository. I see the code displayed in the perldl shell v1.357 (PDL v2.095), but grepping for mandel* finds nothing. I removed it because the code in `demo 3dgal` doesn't use `sever` at all, just `clip` and `lclip`

duffee:
Requested changes made. In addition, I've removed some trailing whitespace, e.g. -> eg on one line that was changing anyways, removed the reference to mandel.pl (noted above) and squashed the commits.

mohawk2:
Thank you!

lib/PDL/Tips.pod
duffee: I found the Mandelbrot code in [PDL::Demos::TriDGallery](https://github.com/PDLPorters/PDL-Graphics-TriD/blob/master/lib/PDL/Demos/TriDGallery.pm) but the code doesn't look anything like the memory savings example in PDL::Tips.

